PR #5, opened 2017-07-31
Add withKnownNat

jmgrosen:
```haskell
withKnownNat :: forall n r. NatRepr n -> (KnownNat n => r) -> r
```

### DIFF
--- a/src/Data/Parameterized/NatRepr.hs
+++ b/src/Data/Parameterized/NatRepr.hs
@@ -32,6 +32,7 @@ module Data.Parameterized.NatRepr
   ( NatRepr
   , natValue
   , knownNat
+  , withKnownNat
   , IsZeroNat(..)
   , isZeroNat
   , NatComparison(..)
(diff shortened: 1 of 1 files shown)


src/Data/Parameterized/NatRepr.hs
```diff
@@ -176,6 +177,14 @@ knownNat = go Proxy
         go p = assert (v <= maxInt) (NatRepr (fromInteger v))
           where v = natVal p
 
+withKnownNat :: forall n r. NatRepr n -> (KnownNat n => r) -> r
+withKnownNat (NatRepr nVal) v =
+  case someNatVal nVal of
+    Just (SomeNat (_ :: Proxy n')) ->
```
robdockins: It probably doesn't make any real difference, but I would feel slightly better if the pattern match also forced the `Proxy` value, just to reinforce that there are no games being played with `undefined`.

jmgrosen: Good point. Updated.

PR #7, opened 2018-03-13
Add precedence support to ShowF

ntc2:
By adding `showsPrecF`, analogous to `showsPrec` in `Show`.

Also, update the `Show` manual instance for `Data.Parameterized.List`
to insert parens when neded, by using `showsPrec` and `showsPrecF`
instead of `show`. E.g. in Brittle

    SynthInstruction RealOpcode AddRr R32 Reg1 :< R32 Reg3 :< Nil

becomes

    SynthInstruction (RealOpcode AddRr) (R32 Reg1 :< (R32 Reg3 :< Nil))

after using `showsPrecF` in place of `showF` in the manual `Show`
instance for `SynthInstruction`.

### DIFF
--- a/src/Data/Parameterized/List.hs
+++ b/src/Data/Parameterized/List.hs
@@ -53,7 +53,9 @@ infixr 5 :<
 instance ShowF f => Show (List f sh) where
   showsPrec _ Nil = showString "Nil"
   showsPrec p (elt :< rest) = showParen (p > precCons) $
-    showsPrecF (precCons+1) elt . showString " :< " . showsPrec (precCons+1) rest
+    -- Unlike a derived 'Show' instance, we don't print parens implied
+    -- by right associativity.
+    showsPrecF (precCons+1) elt . showString " :< " . showsPrec 0 rest
     where
       precCons = 5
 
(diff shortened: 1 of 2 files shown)


ntc2:
There may be other misbehaving `Show` or `ShowF` instances in paremeterized-utils, besides the `List` instance I fixed here. I just fixed the one that I was seeing in Brittle.

src/Data/Parameterized/List.hs
```diff
@@ -51,8 +51,11 @@ data List :: (k -> *) -> [k] -> * where
 infixr 5 :<
 
 instance ShowF f => Show (List f sh) where
-  show Nil = "Nil"
-  show (elt :< rest) = showF elt ++ " :< " ++ show rest
+  showsPrec _ Nil = showString "Nil"
+  showsPrec p (elt :< rest) = showParen (p > precCons) $
+    showsPrecF (precCons+1) elt . showString " :< " . showsPrec (precCons+1) rest
```
joehendrix: Could you modify this to take advantage of right-associativity of ":<" and reduce the number of parens?

To see how Haskell does this for String, see the definition of "showList__" in https://hackage.haskell.org/package/base-4.10.1.0/docs/src/GHC.Show.html.

ntc2: Done. Note that derived `Show` and `Read` instances don't eliminate redundant parens implied by associativity, and derived `Read` instances fail to parse if they are omitted: http://hackage.haskell.org/package/base-4.10.1.0/docs/Prelude.html#t:Show. But of course there are no derived `Read` instances here, so we can do whatever we want.

PR #9, opened 2018-08-23
 Port Vector module from Crucible.Lang 

siddharthist:
- Also port What4.Utils.Endian
- At the suggestion of @robdockins
- All functions mentioning Crucible types have been removed
- This is a prelude to implementing LLVM's bswap intrinsic for Crucible
- Also add Travis CI configuration (build here: https://travis-ci.org/siddharthist/parameterized-utils/jobs/419784177)

### DIFF
--- a/src/Data/Parameterized/Vector.hs
+++ b/src/Data/Parameterized/Vector.hs
@@ -35,6 +35,7 @@ module Data.Parameterized.Vector
 
     -- * Reorder
   , shuffle
+  , reverse
   , rotateL
   , rotateR
   , shiftL
@@ -60,13 +61,19 @@ import Data.Parameterized.NatRepr
 import Data.Proxy
 import Unsafe.Coerce
 import Data.Coerce
-import Prelude hiding (length,zipWith)
+import Prelude hiding (length,reverse,zipWith)
 
 import Data.Parameterized.Utils.Endian
 
 -- | Fixed-size non-empty vectors.
 data Vector n a where
   Vector :: (1 <= n) => !(Vector.Vector a) -> Vector n a
+
+instance Eq a => Eq (Vector n a) where
+  (Vector x) == (Vector y) = x == y
+
+instance Show a => Show (Vector n a) where
+  show (Vector x) = show x
 
 -- | Get the elements of the vector as a list, lowest index first.
 toList :: Vector n a -> [a]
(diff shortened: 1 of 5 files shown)


src/Data/Parameterized/Vector.hs
```diff
@@ -0,0 +1,382 @@
...
+
+-- | Get the elements of the vector as a list, lowest index first.
+toList :: Vector n a -> [a]
+toList (Vector v) = Vector.toList v
+{-# Inline toList #-}
+
+-- | Length of the vector.
+-- @O(1)@
+length :: Vector n a -> NatRepr n
+length (Vector xs) = unsafeCoerce (fromIntegral (Vector.length xs) :: Integer)
```
robdockins: `unsafeCoerce` always gives me heebie-jeebies.  Can we weaken this do `Data.Coerce.coerce`?

Regardless, lets add some comments to this file about why we believe this cast is OK (basically, an invariant that all operations preserve the connection between the type index and the length of the vector.)

src/Data/Parameterized/Vector.hs
```diff
@@ -0,0 +1,382 @@
...
+       fill start (knownNat @0)
+       return v
+
+  inLen :: NatRepr (n * w)
+  inLen = natMultiply n w
+{-# Inline splitWith #-}
+
+
+
+newtype Vec a n = Vec (Vector n a)
```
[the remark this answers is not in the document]
robdockins: Humm, on second thought, probably not... we'd lose the `Functor` instances, etc.   Bummer.  Oh well.

src/Data/Parameterized/Vector.hs
```diff
@@ -0,0 +1,382 @@
...
+    Vector (xs Vector.++ ys)
+  }
+{-# Inline append #-}
+
+
+--------------------------------------------------------------------------------
+
+
+lemmaMul :: (1 <= n) => p w -> q n -> (w + (n-1) * w) :~: (n * w)
+lemmaMul = unsafeCoerce Refl
```
robdockins: Let's move this to `Data.Parameterized.NatRepr`

src/Data/Parameterized/Vector.hs
siddharthist: We don't have to do any coercion at all if we expose the constructor `NatRepr`, but I get a link-time error when I do that. What's going on there?

siddharthist:
@robdockins I've addressed your comments

src/Data/Parameterized/Vector.hs
siddharthist: After a long discussion on Mattermost, I'm leaving this as-is for now. If we want to change this, let's do it in a separate PR.

siddharthist:
@robdockins I've reverted to using `unsafeCoerce` and added a `reverse` function.

siddharthist:
@robdockins I noticed that some of this PR has been merged to master, should I close this?

robdockins:
Yeah, lets close this, I wasn't sure if you were done with it.

siddharthist:
@robdockins If you merge with the Github button, I get a notification, which is nice.

PR #14, opened 2018-10-29
Update for GHC 8.6.1

goldfirere:
This commit should get everything ready for GHC 8.6.1.

There are warnings when building the testsuite, but this seems unchanged from building on 8.4.3, so I did not try to fix.


kquick:
Thank you!

PR #20, opened 2019-01-30
Add a TraversableF instance for Compose'd higher-kinded things

siddharthist:


robdockins:
Looks good to me.

Interesting side note, I investigated adding a `TraverseableFC` instance for compose, but `Compose` has kind `(k -> *) -> (k1 -> k) -> k1 -> *` which fixes `*` as the output kind.  Makes sense, but it wasn't immediately obvious to me that Compose was asymmetric in this way.

PR #27, opened 2019-02-12
Remove unused LANGUAGE pragmas (using hlint)

siddharthist:
Just waiting for CI before merging


TomMD:
Fun fact: Muse can auto run hlint on pull requests.  That's probably going to be more noisy than you'd like though.

siddharthist:
@tommd can it be configured to just do certain checks? This one, for instance, seems unobjectionable, whereas e.g. the warning for extra parentheses is a bit pedantic.

PR #29, opened 2019-04-02
WithRepr & Safe Peano

sweirich:
This PR makes two changes:
  - Adds CPP to the Peano module so that it can be compiled without using unsafeCoerce 
    (not recommended)
  - Adds a new "WithRepr" class that generalizes "withPeanoRepr" and "withKnownNat" 
    (and provides additional instances for Symbol, Assignment and List).

### DIFF
--- a/src/Data/Parameterized/Peano.hs
+++ b/src/Data/Parameterized/Peano.hs
@@ -22,6 +22,7 @@ natural numbers is an Int.
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE PatternGuards #-}
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RoleAnnotations #-}
@@ -44,7 +45,6 @@ module Data.Parameterized.Peano
      , zeroP, succP, predP
 
      , KnownPeano
-     , withKnownPeano
 
      , PeanoRepr, peanoValue
      , PeanoView(..), peanoView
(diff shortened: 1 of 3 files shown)


src/Data/Parameterized/WithRepr.hs
```diff
@@ -0,0 +1,108 @@
...
+data PeanoRepr p where
+    ZRepr :: PeanoRepr Z
+    SRepr :: PeanoRepr p -> PeanoRepr (S p)
+
+-- KnownRepr instances
+@
+
+Then the instance for this class
+@
+instance IsRepr PRepr
```
kquick: Should this be `instance IsRepr PeanoRepr`?

src/Data/Parameterized/Peano.hs
```diff
@@ -151,38 +160,81 @@ data PeanoView (n :: Peano) where
...
 
 instance Hashable (PeanoRepr n) where
-  hashWithSalt i (PeanoRepr x) = hashWithSalt i x
+  hashWithSalt i x = hashWithSalt i (peanoValue x)
 
 instance Eq (PeanoRepr m) where
   _ == _ = True
 
 instance TestEquality PeanoRepr where
+#if UNSAFE_OPS
```
kquick: The cabal file doesn't define a value for UNSAFE_OPS, so these should all be `#ifdef UNSAFE_OPS` instead of `#if UNSAFE_OPS`.

src/Data/Parameterized/WithRepr.hs
sweirich: Yes, thanks. I'll update.

src/Data/Parameterized/Peano.hs
sweirich: Thanks. I've pushed an update.

kquick:
Looks good and I have no more comments.  Please go ahead and merge to master whenever you are ready.

sweirich:
I can’t merge it — I don’t have write access to parameterized-utils.

PR #30, opened 2019-04-05
NatRepr: added Data.Data instance, removed unused Data.Proxy import

benjaminselfridge:
@kquick It would be nice to get this into the hackage release if it is kosher.


kquick:
Looks good: merge to master when you are ready.

PR #32, opened 2019-04-08
Hide activateNatReprCoercionBackdoor from external use

benjaminselfridge:
I'm curious if this change makes sense -- since the backdoor is needed for Vector, perhaps it makes sense to expose it to external uses of parameterized-utils; however, if it's not likely that outside users will ever *need* this function, it might be best to hide it, because it's a pretty nasty wart in the API.


benjaminselfridge:
Yep, sorry about the comments. Could have sworn I committed that, but I guess I hadn't.

joehendrix:
Has there been a problem with people accidentally misusing `activateNatReprCoercionBackdoor_IPromiseIKnowWhatIAmDoing`?  I feel like the function is descriptive enough.  If the problem is the name appearing in the docs, and people not liking it, an alternative solution would be to just delete the function `activateNatReprCoercionBackdoor_IPromiseIKnowWhatIAmDoing`, and replace existing uses with `unsafeCoerce`.  `NatRepr` is just a type around `Natural`, so `unsafeCoerce` should be a fine replacement assuming the previous uses of `activate...` were fine.

benjaminselfridge:
There hasn't been a problem with people using it AFAIK, but it seems better to not include it in the API unless enough people specifically request it. It's a lot easier to add later than it is to remove later.

As for replacing with `unsafeCoerce`, I (and others) like the fact that the name of the function captures intent of use, and think it makes sense to leave it for internal use within `parameterized-utils` rather than just using `unsafeCoerce`.

joehendrix:
I'm against this PR.  I think this adds non-trivial overhead to the process of modifying or extending the `NatRepr` module and doesn't actually provide a benefit that will help users or developers in practice.

A simpler solution would be to just define `NatRepr` in `Data.Parameterized.NatRepr.Internal` and export the constructor.  Then other modules could import `Internal` if they really want to do unsafe things with some slight benefits over `unsafeCoerce`.   To be honest, I'm not sure that provides enough benefit to be worth it, but at least the bulk of the code would still live in `Data.Parameterized.NatRepr` and we wouldn't have to maintain multiple export lists.

I don't particularly care for `activateNatReprCoercionBackdoor_IPromiseIKnowWhatIAmDoing`, but that's for separate reasons: (1) it has a name that is obnoxious to use, (2) the comment doesn't explain what it does and why its useful, and is alarmist, and (3) It's not clear why the CPS style is any better than just exporting a function with the type "Natural -> NatRepr n` with a name like `unsafeNatRepr`.

benjaminselfridge:
Having `Internal.hs` export the constructor seems reasonable enough to me -- Joe, do you feel strongly that users of `parameterized-utils` should be afforded the ability to use that constructor, either via the constructor itself or via something like `unsafeNatRepr`?

If your main qualm is with the redundant export list (which bugged me as well), but you don't mind hiding this functionality from end users, then it seems fine to export the constructor in the internal module, but not expose that module in the `cabal` file.

Do you feel strongly that end users should be able to construct `NatRepr`s from `Natural`s unsafely, or is your opposition more about the awkwardness of redundant export lists?

joehendrix:
My only concern is with the overhead of multiple export lists and need to navigate multiple buffers.  Users can construct `NatRepr`s unsafely no matter what we decide (e.g., `unsafeCoerce`),  I don't have a strong preference on whether `parameterized-utils` exports a somewhat blessed way to do that.

benjaminselfridge:
Ok -- I'll close this PR and take another pass at it I guess.

PR #33, opened 2019-04-08
a few properties about peano arithmetic + correct definition of LE

sweirich:
Sorry about all the PRs.  Just trying to keep things up to date....


src/Data/Parameterized/Peano.hs
```diff
@@ -411,5 +416,51 @@ peanoLength [] = Some zeroP
 peanoLength (_:xs) = case peanoLength xs of
   Some n -> Some (succP n)
 
+
+------------------------------------------------------------------------
+-- Some properties about Peano numbers
+
+plusCtxSizeAxiom :: forall t1 t2 f.
```
kquick: Can you add some quick haddocks for these since they are exported?

sweirich: Done.

kquick:
Thanks, Stephanie!

PR #34, opened 2019-04-08
created a separate module, Data.Parameterized.NatRepr.Internal, which…

benjaminselfridge:
… contains the

definition of NatRepr and everything required for its instances, and exports everything by default. This module is not exposed to the end user, but is needed within Data.Parameterized.Vector.

This should eliminate the need for maintaining two import lists, and doesn't provide a "blessed" way to coerce 'Naturals' into 'NatRepr's.

### DIFF
--- a/parameterized-utils.cabal
+++ b/parameterized-utils.cabal
@@ -78,6 +78,9 @@ library
     Data.Parameterized.Vector
     Data.Parameterized.WithRepr                  
 
+  other-modules:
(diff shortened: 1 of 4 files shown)


kquick:
I like this better and I'm OK with merging this, pending comments from Joe or Rob.

joehendrix:
I'm still skeptical of the value, but this is much cleaner than the last effort.  I'm ok if other people think it's important.

PR #36, opened 2019-04-15
Tr/documentation

travitch:
I'd like to add some more documentation to parameterized-utils.  This is a start.  The change to Data.Parameterized.Context.{Safe,Unsafe} might be controversial, but hopefully the README is not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+# Overview
+
 The parameterized-utils module contains a collection of typeclasses
 and datatypes for working with parameterized types, that is types that
 have a type argument.  One example would be a algebraic data type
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,7 +1,246 @@
...
+
+* Data.Parameterized.Map (`MapF (key :: k -> Type) (value :: k -> Type)`)
+
+  `MapF` an associative map from keys to values where both keys and values are
+  parameterized types.  The lookup operation is O(log(n)), and recovers the type
+  parameter of the value during lookup.
+
+* Data.Parameterized.HashTable (`HashTable s (key :: k -> Type) (value :: k -> Type)`)
+
+  `HashTable` is an associative container like ``MapF`, except is mutable in
```
kquick: Duplicate backslash causes switch to the wrong mode here.

README.md
```diff
@@ -1,7 +1,246 @@
...
+
+  This module provides run-time representative values for booleans lifted to
+  the type level.
+
+* Data.Parameterized.Some
+
+  The `Some` type is a wrapper that existentially quantifies away the type
+  parameter of a parameterized value.  This can be used on any value with a
+  parameterized type, but is most useful when an operation exists to recover the
+  type parameter later (either via pattern matching over a GADT or by consulting
```
kquick: open paren with no closure

README.md
```diff
@@ -1,7 +1,246 @@
...
+  a run-time type representative value.
+
+* Data.Parameterized.Nonce
+
+  `Nonce` is a parameterized type backed by a `Word64`.  Its `TestEquality`
+  instance uses `unsafeCoerce` to allow the type parameter to be recovered.
+  Similarly to a cryptographic nonce, the `Nonce` type is safe as long as no
+  nonce value is reused.
```
kquick: Maybe describe why a Nonce would be preferred to just using a Word64.

README.md
```diff
@@ -1,7 +1,246 @@
...
+  `Assignment` is a sequence type that holds values of parameterized types.
+  It is essentially a snoc list (i.e., a list that is extended on the right
+  instead of the left).  In the default implementation, indexing is O(log(n))
+  time and total.
+
+* Data.Parameterized.List (`List (f :: k -> Type) [k]`)
+
+  `List` is the plain Haskell list lifted to hold values of parameterized
+  types.  Moreover, it uses the data kind lifted list syntax instead of the
+  `Ctx` type.  Indexing into `List` is total but O(n).
```
kquick: The `Ctx` data structure is never described, just referred to here and used in the `Context` above.

README.md
```diff
@@ -1,7 +1,246 @@
...
+parameterized types or otherwise type indexed.  The following modules implement
+data structures:
+
+* Data.Parameterized.Context (`Assignment (f :: k -> Type) (ctx :: Ctx k)`)
+
+  `Assignment` is a sequence type that holds values of parameterized types.
+  It is essentially a snoc list (i.e., a list that is extended on the right
+  instead of the left).  In the default implementation, indexing is O(log(n))
+  time and total.
+
```
kquick: Hiding the Safe/Unsafe in the Haddocks seems generally fine (although it prevents viewing the source to see the implementation differences), but the discussion here should probably talk about the Safe/Unsafe implementation distinctions, esp. wrt "the default implementation" teaser above.

PR #49, opened 2019-06-13
New module: All

siddharthist:
This is useful e.g. when defining instances for only polymorphic instances of an indexed datatype. 

For a concrete example of how I'd use this in SAW, I'd like to add `IsValue`/`FromValue` instances to `Any CrucibleSetupValue`, i.e. only those `SetupValue`s that are polymorphic in the LLVM architecture.

### DIFF
--- a/src/Data/Parameterized/All.hs
+++ b/src/Data/Parameterized/All.hs
@@ -7,11 +7,33 @@
 --
 -- This module provides 'All', a GADT that encodes universal
 -- quantification/parametricity over a type variable.
+--
+-- The following is an example of a situation in which it might be necessary
+-- to use 'All' (though it is a bit contrived):
+--
+-- @
+--   {-# LANGUAGE FlexibleInstances #-}
+--   {-# LANGUAGE GADTs #-}
+--
+--   data F (x :: Bool) where
+--     FTrue :: F 'True
+--     FFalse :: F 'False
+--     FIndeterminate :: F b
+--
+--   data Value =
+--     VAllF (All F)
+--
+--   class Valuable a where
+--     valuation :: a -> Value
+--
+--   instance Valuable (All F) where
+--     valuation = VAllF
+--
+--   val1 :: Value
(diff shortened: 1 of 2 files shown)


benjaminselfridge:
Cool!

src/Data/Parameterized/All.hs
```diff
@@ -0,0 +1,40 @@
...
+
+import Data.Functor.Const (Const(..))
+
+import Data.Parameterized.Classes
+import Data.Parameterized.TraversableF
+
+newtype All (f :: k -> *) = All { getAll :: forall x. f x }
+
+instance FunctorF All where
+  fmapF nat (All a) = All (nat a)
```
kquick: Use 'f' instead of 'nat' here?

src/Data/Parameterized/All.hs
```diff
@@ -0,0 +1,40 @@
+------------------------------------------------------------------------
+-- |
+-- Module           : Data.Parameterized.All
+-- Copyright        : (c) Galois, Inc 2014-2019
```
kquick: Probably just 2019 for this module.  :-)

kquick:
Since this is introducing a new module, can you provide a description and motivating example that is standalone (e.g. not a reference to crucible) that can be used for Haddock and other documentation?

siddharthist:
Here's an example which is just the bare-bones of the situation I'm actually dealing with, I'm not sure it's sufficiently "motivating" though (it seems a bit contrived)
```haskell
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE GADTs #-}

data F (x :: k) where
  FUnit :: F ()
  FInt :: F Int

data Value =
  VAllF (All F)

class Valuable a where
  valuation :: a -> Value

instance Valuable (All F) where
  valuation = VAllF
```

siddharthist:
@kquick I have added my contrived example to the module documentation, I'll keep thinking about a simpler one.

PR #50, opened 2019-06-13
Implement IndexView

glguy:
This allows us to deconstruct indexes in a typed way

### DIFF
--- a/src/Data/Parameterized/Context/Unsafe.hs
+++ b/src/Data/Parameterized/Context/Unsafe.hs
@@ -13,6 +13,7 @@
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE TypeInType #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE StandaloneDeriving #-}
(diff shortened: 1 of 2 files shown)


src/Data/Parameterized/Context/Unsafe.hs
```diff
@@ -294,6 +299,24 @@ instance Show (Index ctx tp) where
 
 instance ShowF (Index ctx)
 
+-- | View of indexes as pointing to the last element in the
+-- index range or pointing to an earlier element in a smaller
+-- range.
+data IndexView ctx tp where
+  IndexViewHere :: !(Size ctx) -> IndexView (ctx '::> tp) tp
+  IndexViewThere :: !(Index ctx tp) -> IndexView (ctx '::> u) tp
```
joehendrix: This is bike shedding, and the changes look fine.   However, what if we called the constructs `IndexViewLast` and `IndexViewInit` respectively.  They share some similarity with `last` and `init` in `Data.List`.

glguy: I'm on board

PR #55, opened 2019-07-26
traverseWithIndex_ for mapping functions returning ().

benjaminselfridge:
I know it's easy to get from `traverseAndCollect`, but it might make sense to include this.

### DIFF
--- a/src/Data/Parameterized/Context.hs
+++ b/src/Data/Parameterized/Context.hs
@@ -52,6 +52,7 @@ module Data.Parameterized.Context
   , generateSomeM
   , fromList
   , traverseAndCollect
+  , traverseWithIndex_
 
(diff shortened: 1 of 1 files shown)


src/Data/Parameterized/Context.hs
```diff
@@ -381,6 +383,14 @@ traverseAndCollect ::
 traverseAndCollect f =
   runCollector . traverseWithIndex (\i x -> Collector (f i x))
 
+-- | Visit each of the elements in an @Assignment in order
```
siddharthist:
```suggestion
-- | Visit each of the elements in an @Assignment@ in order
```

kquick:
This looks good, @benjaminselfridge , were you ready to merge this?

PR #56, opened 2019-08-06
Introduce ascending and descending sorts to map.

joehendrix:


kquick:
These changes look fine, were you ready to merge this, @joehendrix ?

PR #61, opened 2019-08-30
Switch to use travis over cabal.

joehendrix:
This switches the travis build to use stack instead of cabal.

There are four immediate motivations for this:
1. I'd like the build process to run with `-Wall and -Werror` so that we fix warnings in `parameterized-utils`.
2. I'd like to have an easy way of pinning package versions for repeatable builds.
3. I'd like to build with GHC 8.6.
4. I'd prefer our travis script to be simple enough that we do not need a tool to generate it.

We don't necessarily need to use `stack`, but points 1 did not seem to be easy to do in `cabal new-build` as passing `--ghc-options=-Wall -Werror` to cabal new-build resulted in all packages building with those flags, and failing whereas I only want parameterized-utils building with those flags.  I also prefer stockage version pinning over cabal freeze, as I like being able to reference a version number.


joehendrix:
The `cabal.project` approach for (1) seems reasonable to me, but we'd probably want a `cabal.project.travis` file so users can have their own, and it's not clear to me how to auto-generate then patch the travis script (or maybe `haskell-ci` has some interface for this)?

I'm going to merge this since it does cache the stack build artifacts, and we can easily tweak it as needed for other checks if linting is desired, though I'm not quite sure what building from sdist entails.

PR #64, opened 2019-09-11
Update parameterized-utils.cabal

glguy:
There are new versions of this package's dependencies to enable parameterized-utils to build on ghc 8.8.1


kquick:
Looks good, thanks for the updates, @glguy.

PR #68, opened 2020-01-06
Add the `splitPrefix` operation, which splits an assignment

robdockins:
into a prefix and all remaining arguments, provided the
assignment begins with the given prefix value.


robdockins:
I think it would be difficult to unify these code paths in this case, because `take` already knows it will succeed before you start (because of the structure of the types), whereas `splitPrefix` has to perform a test.  Maybe there is a more general operation of some sort, but it would be pretty tricky, I think.

PR #72, opened 2020-03-13
Revert to cuckoo implementation for HashTable.

kquick:
This is *not* intended to be merged to master: this is a test to check
if the change from cuckoo to linear caused a performance regression.


kquick:
Performance testing completed, this branch is no longer needed.

PR #73, opened 2020-03-13
Change HashTable algorithm to "Basic".

kquick:


kquick:
This addresses a performance regression in various SAW-script tests.  Tests that the Cuckoo HashTable implementation could complete in 28 minutes take 88m with the Basic algorithm, but they take 10+ hours with the Linear algorithm.

joehendrix:
So that it is captured somewhere, what is the bug that prevents using the Cuckoo implementation?

kquick:
The issues with the cuckoo implementation were identified by @benjaminselfridge and @travitch ; I will have one of them provide updates here.

travitch:
@benjaminselfridge noticed some cases where the hash table was not returning values that has been inserted. In our case, it led to getting spurious `undefined` values out of the table. I suspect that this is the issue: gregorycollins/hashtables#54.  That is one of two pretty serious bugs open on the Cuckoo hash table that have been open for almost a year, so our feeling was that a fix is not imminent.  I'd love to switch back when we can.

joehendrix:
Thanks -- that's useful to know.

PR #78, opened 2020-06-24
add CtxFlatten type family and flattenAssignment

Ptival:
For some purposes, it is useful to forget the nested structure of a
context of contexts, flattening it into one large context.  This
commit introduces a type family CtxFlatten to capture this at the
type level.

At the value level, the relevant operation consists in flattening
an assignment of assignments over this context of contexts, into
a flat assignment.

### DIFF
--- a/src/Data/Parameterized/Context.hs
+++ b/src/Data/Parameterized/Context.hs
@@ -65,6 +65,8 @@ module Data.Parameterized.Context
   , traverseWithIndex_
   , dropPrefix
   , unzip
+  , flattenAssignment
+  , flattenSize
 
     -- * Context extension and embedding utilities
   , CtxEmbedding(..)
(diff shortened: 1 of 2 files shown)


src/Data/Parameterized/Context.hs
```diff
@@ -185,6 +187,26 @@ dropPrefix xs0 prefix err = go xs0 (sizeInt (size xs0))
...
+  Assignment f (CtxFlatten ctxs)
+flattenAssignment ctxs =
+  case viewAssign ctxs of
+    AssignEmpty -> empty
+    AssignExtend ctxs' ctx -> flattenAssignment ctxs' <++> ctx
+
+-- | Given the size of each context in @ctxs@, returns the size of @CtxFlatten
+-- ctxs@.  You can obtain the former from any nested assignment @Assignment
+-- (Assignment f) ctxs@, by calling @fmapFC size@.
+flattenSize ::
```
robdockins, line 203: This isn't building because of the `Ctx` import prefix.  Looks like you copied this from somewhere else and didn't clean it up.

Ptival: Indeed, fixed in 6dc5c48b3aff5c1dd86035d445997dea3d4c7646

robdockins (approved):

PR #81, opened 2020-07-20
Re-organize with internal libraries to hide safe/unsafe internals.

kquick:
This hides the Safe/Unsafe implementations of Context which prevents a
client from improperly using the wrong one.

This also allows client code and tests to be written in a fully
generic manner.

The internal movement/splitting of files is somewhat unfortunate, but
Cabal will look for a local file to supply an import statement before
checking libraries, so the files had to be segregated physically as
well as by internal library build.


langston-barrett:
@kquick What's the status of this PR?

langston-barrett:
Closing since it's been 5 years and there's a lot of merge conflicts.

PR #92, opened 2021-01-15
Add SomeSym to hide Symbol parameters while preserving KnownSymbol.

kquick:


robdockins (approved):
Seems OK, except that CI indicates it doesn't build on GHC 8.4.  I think you need to `import Data.Kind (Type)`

travitch:
It seems like neither the type nor the function are exported? If it is exported, should we make the actual constructor name match the type name?

PR #93, opened 2021-01-15
Remove stack YAML files since they are probably out-of-date.

kquick:


robdockins (approved):

kquick:
Actually this needs to wait until I update the CI, since the Travis currently uses these.

PR #96, opened 2021-02-06
Conversions between Assignment and Vector

langston-barrett:


kquick (approved):
Thanks for the tests!

My only concern is whether the Vector order is what is desired, since Assignment is typically in reverse order from List.  Regardless of which way the resulting Vector entries should be considered, you probably want to add that to the documentation for the `fromAssignment` and `toAssigment` haddocks.

langston-barrett:
> My only concern is whether the Vector order is what is desired, since Assignment is typically in reverse order from List.

Good point. The order here is determined by the pre-existing `Context.toVector` function:

https://github.com/GaloisInc/parameterized-utils/blob/e7b080b82a1d6e73cb6d3e5376e88dbca625faba/src/Data/Parameterized/Context.hs#L151-L157

I'll document the order on that function, and then note in the Haddock for `fromAssignment` that it uses the same order as `toVector.

> Regardless of which way the resulting Vector entries should be considered, you probably want to add that to the documentation for the fromAssignment and toAssigment haddocks.

Will do, thanks!

PR #97, opened 2021-02-17
Context: Add unzip operation for assignments of products

langston-barrett:


kquick (approved):
Looks pretty straightforward, but would you be able to add a test case for this?

langston-barrett:
> would you be able to add a test case for this?

Done!

PR #100, opened 2021-03-15
Misc comment updates.

kquick:


src/Data/Parameterized/Context/Safe.hs
```diff
@@ -388,7 +388,7 @@ forIndexRangeImpl i (SizeSucc sz) d f r =
 
 -- | Given an index @i@, size @n@, a function @f@, value @v@, and a
 -- function @f@, the expression @forIndexRange i n f v@ is equivalent
```
RyanGlScott, line 390: This isn't your fault, but I noticed that this redundantly mentions "`a function @f@`" a second time. Let's delete the second one.

src/Data/Parameterized/Context/Unsafe.hs
```diff
@@ -315,7 +315,7 @@ forIndex n f r =
 
 -- | Given an index @i@, size @n@, a function @f@, value @v@, and a
 -- function @f@, the expression @forIndex i n f v@ is equivalent to
```
RyanGlScott, line 317: Again, "`a function @f@`" is redundantly mentioned twice here.

RyanGlScott (approved):
LGTM aside from a couple of minor suggestions.

PR #101, opened 2021-03-15
Update cabal specification and validation.

kquick:


travitch (changes requested):
The metadata updates are all great.

I'm a bit concerned about the nix build changes. I understand that the goal there is to ensure that the version of cabal requested in the cabal file is consistent with the version that GHC is linked against (i.e., that we aren't relying on a version of cabal too new for older GHCs). I'd much prefer to see this concern addressed in the Haskell build action in the long term, as that would support MacOS and Windows. I also have some concerns around the stability of the nixos package channels. Maybe more importantly, I'm not sure that others will be able to effectively maintain the nix bits.

Can we switch the build back to using the github action, but leave the cabal version check using the nix tools?

travitch (approved):

PR #103, opened 2021-03-27
WIP: A solver for type-level equations in monoids

langston-barrett:
Needs a lot of cleanup/work on the interface. See the paper for inspiration and applications: https://webspace.science.uu.nl/~swier004/publications/2021-jfp-submission.pdf

### DIFF
--- a/src/Data/Parameterized/MonoidSolver.hs
+++ b/src/Data/Parameterized/MonoidSolver.hs
@@ -31,11 +31,8 @@ import Data.Parameterized.NatRepr (type (+), plusAssoc)
 
 -- | 'Op' corresponds to @Add@ from the paper.
 type family Op k (m1 :: k) (m2 :: k) :: k
--- | 'Unit' corresponds to @Add@ from the paper.
+-- | 'Unit' corresponds to @Zero@ from the paper.
 type family Unit k :: k
-
-type instance Op Nat n1 n2 = n1 + n2
-type instance Unit Nat = 0
 
 -- | Type level only
 data MonoidExpr k
@@ -98,13 +95,13 @@ type family Normalize (me :: MonoidExpr m) :: MonoidExpr m where
 -- subexpressions of its argument.
 normalizeLemma ::
   forall k me1 me2.
-  (forall proxy n. proxy n -> Proxy (Op k (Unit k) n) :~: Proxy n) ->
-  (forall proxy n. proxy n -> Proxy (Op k n (Unit k)) :~: Proxy n) ->
-  (forall proxy n m l. proxy n -> proxy m -> proxy l -> Proxy (Op k n (Op k m l)) :~: Proxy (Op k (Op k n m) l)) ->
+  (forall proxy n. proxy n -> Op k (Unit k) n :~: n) ->
+  (forall proxy n. proxy n -> Op k n (Unit k) :~: n) ->
+  (forall proxy n m l. proxy n -> proxy m -> proxy l -> Op k n (Op k m l) :~: Op k (Op k n m) l) ->
   MERepr k me1 ->
   MERepr k me2 ->
-  Proxy (Eval k (Apply (Diff me1) (Apply (Diff me2) 'EUnit))) :~:
-    Proxy (Op k (Eval k me1) (Eval k me2))
+  Eval k (Apply (Diff me1) (Apply (Diff me2) 'EUnit)) :~:
+    Op k (Eval k me1) (Eval k me2)
 normalizeLemma idl idr assoc mer1 mer2 =
   case mer1 of
     EUnitRepr ->
@@ -120,15 +117,15 @@ normalizeLemma idl idr assoc mer1 mer2 =
         Refl ->
               assoc (Proxy :: Proxy (Eval k me1')) (Proxy :: Proxy (Eval k me2')) (Proxy :: Proxy (Eval k me2))
   where
-    norm :: MERepr k me -> Proxy (Eval k (Normalize me)) :~: Proxy (Eval k me)
+    norm :: MERepr k me -> Eval k (Normalize me) :~: Eval k me
     norm = normalizeSound idl idr assoc
 
 normalizeSound ::
-  (forall proxy n. proxy n -> Proxy (Op k (Unit k) n) :~: Proxy n) ->
-  (forall proxy n. proxy n -> Proxy (Op k n (Unit k)) :~: Proxy n) ->
-  (forall proxy n m l. proxy n -> proxy m -> proxy l -> Proxy (Op k n (Op k m l)) :~: Proxy (Op k (Op k n m) l)) ->
+  (forall proxy n. proxy n -> Op k (Unit k) n :~: n) ->
+  (forall proxy n. proxy n -> Op k n (Unit k) :~: n) ->
+  (forall proxy n m l. proxy n -> proxy m -> proxy l -> Op k n (Op k m l) :~: Op k (Op k n m) l) ->
   MERepr k me ->
-  Proxy (Eval k (Normalize me)) :~: Proxy (Eval k me)
+  Eval k (Normalize me) :~: Eval k me
 normalizeSound idl idr assoc =
   \case
     EUnitRepr -> Refl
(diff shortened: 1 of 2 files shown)


src/Data/Parameterized/MonoidSolver.hs
```diff
@@ -0,0 +1,210 @@
...
+  Eval k ('EOp me1 me2) = Op k (Eval k me1) (Eval k me2)
+
+-- | Defunctionalization - inspired by \"singletons\", and with a similar naming
+-- scheme
+data TyFun a b
+type a ~> b = TyFun a b -> Type
+type family Apply (f :: a ~> b) (x :: a) :: b
+
+data IdSym1 :: MonoidExpr m ~> MonoidExpr m
+data EOpSym0 :: MonoidExpr m ~> (MonoidExpr m ~> MonoidExpr m)
```
RyanGlScott, line 66: If you declare `infixr 0 ~>`, then the parentheses around `(MonoidExpr m ~> MonoidExpr m)` become redundant. Similar comments apply to other defunctionalization symbols below.

src/Data/Parameterized/MonoidSolver.hs
```diff
@@ -0,0 +1,210 @@
...
+data EOpSym1 (me :: MonoidExpr m) :: MonoidExpr m ~> MonoidExpr m
```
RyanGlScott, line 67: This ultimately comes down to personal taste, but I would prefer to write this as:

```suggestion
data EOpSym1 :: MonoidExpr m -> MonoidExpr m ~> MonoidExpr m
```

And similarly for other defunctionalization symbols that require at least one argument.

src/Data/Parameterized/MonoidSolver.hs
```diff
@@ -0,0 +1,210 @@
...
+
+-- | 'UnDiff' corresponds to "reify" from the paper.
+type family UnDiff (diff :: MonoidExpr m ~> MonoidExpr m) :: MonoidExpr m where
+  UnDiff diff = Apply diff 'EUnit
+
+type family Normalize (me :: MonoidExpr m) :: MonoidExpr m where
+  Normalize me = UnDiff (Diff me)
+
+-- | Note on termination: The one recursive call to this function is strictly
+-- decreasing on its first argument. This function is mutually recursive with
```
RyanGlScott, line 95: By "first argument", do you mean `mer1`? I ask since that appears to be the fourth argument overall to this function.

src/Data/Parameterized/MonoidSolver.hs
```diff
@@ -0,0 +1,210 @@
...
+-- 'normalizeSound' and recursive calls to that function are made on the second
```
RyanGlScott, line 96: Again, I'm not sure if "second argument" is strictly accurate here.

src/Data/Parameterized/MonoidSolver.hs
```diff
@@ -0,0 +1,210 @@
...
+solveNat =
+  solve
+    (const Refl)
+    (const Refl)
+    (\proxy1 proxy2 proxy3 ->
+       case plusAssoc proxy1 proxy2 proxy3 of
+         Refl -> Refl)
+
+----------------------------------------------------------------------
+-- Examples
```
RyanGlScott, line 165: These examples would be a better fit in the test directory, I think.

src/Data/Parameterized/MonoidSolver.hs
```diff
@@ -0,0 +1,210 @@
+{-|
+Description : A solver for type-level equations in monoids
+Copyright   : (c) Galois, Inc 2021
+Maintainer  : Langston Barrett
+
+Implementation of section 6 of \"A well-known representation of monoids and its
+application to the function \'vector reverse\'\" by Wouter Swierstra.
```
RyanGlScott, line 7: Include a link to the paper here?

src/Data/Parameterized/MonoidSolver.hs
```diff
@@ -0,0 +1,210 @@
...
+  Normalize me = UnDiff (Diff me)
+
+-- | Note on termination: The one recursive call to this function is strictly
+-- decreasing on its first argument. This function is mutually recursive with
+-- 'normalizeSound' and recursive calls to that function are made on the second
+-- argument, but calls to this function *from* 'normalizeSound' are on strict
+-- subexpressions of its argument.
+normalizeLemma ::
+  forall k f me1 me2.
+  (forall proxy n. proxy n -> f (Op k (Unit k) n) :~: f n) ->
```
RyanGlScott, line 101: This (and the following two arguments) are repeated often enough that I wonder if it's worth declaring type synonyms for them. Something like:

```hs
type UnitLeft k (f :: k -> Type) =
  forall (proxy :: k -> Type) (n :: k).
  proxy n -> f (Op k (Unit k) n) :~: f n

type UnitRight k (f :: k -> Type) =
  forall proxy n.
  proxy n -> f (Op k n (Unit k)) :~: f n

type OpAssoc k (f :: k -> Type) =
  forall (proxy :: k -> Type) (n :: k) (m :: k) (l :: k).
  proxy n -> proxy m -> proxy l -> f (Op k n (Op k m l)) :~: f (Op k (Op k n m) l)

normalizeLemma ::
  forall k f me1 me2.
  UnitLeft k f ->
  UnitRight k f ->
  OpAssoc k f ->
  MERepr k f me1 ->
  MERepr k f me2 ->
  f (Eval k (Apply (Diff me1) (Apply (Diff me2) 'EUnit))) :~:
    f (Op k (Eval k me1) (Eval k me2))

normalizeSound ::
  UnitLeft k f ->
  UnitRight k f ->
  OpAssoc k f ->
  MERepr k f me ->
  f (Eval k (Normalize me)) :~: f (Eval k me)

solve ::
  UnitLeft k f ->
  UnitRight k f ->
  OpAssoc k f ->
  MERepr k f me1 ->
  MERepr k f me2 ->
  f (Eval k (Normalize me1)) :~: f (Eval k (Normalize me2)) ->
  f (Eval k me1) :~: f (Eval k me2)
```

src/Data/Parameterized/MonoidSolver.hs
langston-barrett: Will do!

src/Data/Parameterized/MonoidSolver.hs
langston-barrett: Good to know, thanks!

src/Data/Parameterized/MonoidSolver.hs
langston-barrett: Any particular reason/principle? Happy to change it either way, doesn't make a difference to me.

src/Data/Parameterized/MonoidSolver.hs
langston-barrett: After packaging the axioms into a typeclass, this is accurate! Good catch :smile:

src/Data/Parameterized/MonoidSolver.hs
langston-barrett: Like the above, this is fixed by packaging the axioms into a typeclass.

src/Data/Parameterized/MonoidSolver.hs
langston-barrett: Agreed, this is part of why this is still WIP :-)

src/Data/Parameterized/MonoidSolver.hs
langston-barrett: Fixed, in the latest draft the axioms are packaged as a type family.

src/Data/Parameterized/MonoidSolver.hs
RyanGlScott: Mainly because it makes the differences between the kinds of each defunctionalization symbol more apparent:

```hs
data EOpSym0 :: MonoidExpr m ~> MonoidExpr m ~> MonoidExpr m
data EOpSym1 :: MonoidExpr m -> MonoidExpr m ~> MonoidExpr m
```

Here, the difference between `EOpSym0` and `EOpSym1` is which arrow kind is used for the first argument.

src/Data/Parameterized/MonoidSolver.hs
```diff
@@ -0,0 +1,283 @@
...
+  ( KnownRepr (MERepr k) me1
+  , KnownRepr (MERepr k) me2
+  ) => KnownRepr (MERepr k) (me1 ':<>: me2) where
+  knownRepr = EOpRepr knownRepr knownRepr
+
+----------------------------------------------------------------------
+-- Instances
+--
+
+type instance Op Nat n1 n2 = n1 + n2
```
RyanGlScott, line 196: Now that you have a `TypeLevelMonoid` class, would it make more sense to have `Op` and `Unit` be associated type families in `TypeLevelMonoid`? Unless I'm missing something, any time you'd define instances for `Op` and `Unit`, you'd also want to define a corresponding instance for `TypeLevelMonoid` as well, so better to make this relationship explicit.

langston-barrett: Yes, good point!

benjaminselfridge:
Holy crap, this is so cool!

langston-barrett:
I don't think we have enough need for this within Galois to justify finishing and maintaining it.

PR #106, opened 2021-04-14
Add `withLeqTrans` to `NatRepr`

Ptival:
Would such a helper be considered for inclusion in `Data.Parameterized.NatRepr`?

In `GRAFT`, we have a lot of code that calls `bv-sized` functions that incur constraints like `3 <= w`, `5 <= w`, `13 <= w`, etc. while working in an environment where we know `32 <= w` or `64 <= w`.

GHC does not automatically dispatch such trivial transitive inequality constraints, and the plug-ins that do such type-level reasoning are experimental and scary to depend upon. This allows wrapping the code in a harness:

```haskell
foo :: forall w. KnownNat w => 64 <= w => a
foo =
  withLeqTrans (knownNat @3) (knownNat @64) (knownNat @w) $ 
  withLeqTrans (knownNat @5) (knownNat @64) (knownNat @w) $ 
  withLeqTrans (knownNat @13) (knownNat @64) (knownNat @w) $
  originalCode -- needs 3 <= w, 5 <= w, 13 <= w
```

which, while verbose, is easy and safe.

I'm open to suggestions!

### DIFF
--- a/src/Data/Parameterized/NatRepr.hs
+++ b/src/Data/Parameterized/NatRepr.hs
@@ -87,6 +87,7 @@ module Data.Parameterized.NatRepr
   , testStrictLeq
   , leqRefl
   , leqTrans
+  , withLeqTrans
   , leqAdd2
   , leqSub2
   , leqMulCongr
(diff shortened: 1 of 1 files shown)


Ptival:
Hmm actually I just realized there's a `withLeqProof` and the whole thing is easily implemented as:
```haskell
withLeqProof (leqTrans (leqProof small middle) (leqProof middle large))
```

PR #107, opened 2021-04-15
Fixes (I think) a slight error in the laws for OrdF

benjaminselfridge:
When the two sides have the same `k tp`, I think we want `compareF` to respect
`Ord` in all cases.


benjaminselfridge:
@robdockins Does this make sense?

robdockins (approved):
Yeah, looks to me like the previous comment wasn't quite right.

PR #108, opened 2021-04-23
Adds some instances for Proxy.

benjaminselfridge:
@robdockins Do these seem reasonable?


travitch (commented):
This seems reasonable to me. I'd say that there is another strictness choice for the `EqF` instance (i.e., don't force the two values). I don't know what the other instances of `EqF` do. I've also never seen `EqF` used anywhere, so it might not be very important.

travitch (approved):

PR #109, opened 2021-04-24
Showf type synonym

benjaminselfridge:
This replaces the `ShowF` class with a type synonym:

```
type ShowF f = forall tp . Show (f tp)
```

I don't know if we want this change, but I've made the PR to have a pros/cons discussion about it.

Pros
* Reduces boilerplate (no more `instance ShowF MyType`)
* For types defined elsewhere, we need orphan `instance ShowF` to get a `Show` instance on `List`, `Assignment`, etc. This makes that no longer necessary

Cons
* Drops support for GHC 8.4
* Have to define `instance Show (MyType a)`, rather than individual `instance Show (MyType Foo)`, `instance Show (MyType Bar)`

### DIFF
--- a/src/Data/Parameterized/All.hs
+++ b/src/Data/Parameterized/All.hs
@@ -38,6 +38,7 @@
 ------------------------------------------------------------------------
 
 {-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE QuantifiedConstraints #-}
 {-# LANGUAGE RankNTypes #-}
 
 module Data.Parameterized.All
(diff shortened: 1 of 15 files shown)


travitch:
If we were going to do something like this, I'd really prefer to wait until we have the automation in place to test across all known affected repositories

benjaminselfridge:
@travitch That makes sense. I know fryingpan is down, but once it's back up would that be sufficient?

This may easily be an instance of "if it ain't broke, don't fix it." But the orphan instance thing has become slightly annoying for using `Data.Parameterized` containers on GADTs defined in other libraries.

travitch:
Yes, I was referring to fryingpan. I like this idea and would love to get rid of these instances, but would really like to avoid any possibility of us getting 80% of the way to done (with some PRs merged) and then finding out that something downstream cannot use this approach for some reason that nobody foresaw.

benjaminselfridge:
Great, then I'll wait until fryingpan is back up before pushing this forward. Thanks!

langston-barrett:
We now have Github actions on many repos that might be affected (What4, Crucible, Macaw, SAW). It would be helpful to make branches on at least a few of those repos that demonstrate converting to this new interface.

PR #111, opened 2021-05-08
Adds TH primitives for generating reprs

benjaminselfridge:
`mkRepr` generates runtime repr types for a given data kind (just like the singletons library). `mkKnownRepr `generates `KnownRepr` instances for data kinds.

### DIFF
--- a/src/Data/Parameterized/TH/GADT.hs
+++ b/src/Data/Parameterized/TH/GADT.hs
@@ -715,9 +715,9 @@ getCtorName c = case c of
 --
 -- Assuming the user of @parameterized-utils@ follows this convention, we
 -- provide the Template Haskell construct 'mkRepr' to automate the creation of
--- the @TRepr@ GADT type. We also provide 'mkKnownReprs', which generates
--- 'KnownRepr' instances for that GADT type. See the documentation for those two
--- functions for more detailed explanations.
+-- the @TRepr@ GADT. We also provide 'mkKnownReprs', which generates 'KnownRepr'
+-- instances for that GADT type. See the documentation for those two functions
+-- for more detailed explanations.
 --
 -- NB: These macros are inspired by the corresponding macros provided by
 -- `singletons-th`, and the "repr" programming idiom is very similar to the one
--- a/src/Data/Parameterized/TH/GADT.hs
+++ b/src/Data/Parameterized/TH/GADT.hs
@@ -718,3 +718,7 @@ getCtorName c = case c of
 -- the @TRepr@ GADT type. We also provide 'mkKnownReprs', which generates
 -- 'KnownRepr' instances for that GADT type. See the documentation for those two
 -- functions for more detailed explanations.
+--
+-- NB: These macros are inspired by the corresponding macros provided by
+-- `singletons-th`, and the "repr" programming idiom is very similar to the one
+-- used by `singletons.`
--- a/src/Data/Parameterized/TH/GADT.hs
+++ b/src/Data/Parameterized/TH/GADT.hs
@@ -546,6 +546,17 @@ matchShowCtor p con = showCon p (constructorName con) (length (constructorFields
 --   where
 --     T4Repr :: NatRepr tp1 -> PeanoRepr tp2 -> T4Repr ('T4 tp1 tp2)
 -- @
+--
+-- Currently, only monomorphic data kinds are supported, so the following will not work:
+--
+-- @
+-- data T5 a = T5 a
+-- \$(mkRepr ''T5)
+-- ======>
+-- Foo.hs:1:1: error:
+--     Exception when trying to run compile-time code:
+--       mkRepr cannot be used on polymorphic data kinds.
+-- @
 mkRepr :: Name -> DecsQ
 mkRepr typeName = do
   let reprTypeName = mkReprName typeName
@@ -575,6 +586,10 @@ mkRepr typeName = do
              ctors
              []
            ]
+  where getCtorName :: Type -> Name
+        getCtorName c = case c of
+          ConT nm -> nm
+          _ -> error $ "mkRepr cannot be used on polymorphic data kinds."
 
 -- | Generate @KnownRepr@ instances for each constructor of a data kind. Given a
 -- data kind @T@, we assume a repr type @TRepr (t :: T)@ is in scope with
@@ -631,14 +646,13 @@ mkKnownReprs typeName = do
         krDec = FunD krFName [Clause [] (NormalB krExp) []]
 
     return $ InstanceD Nothing krReqs krConstraint [krDec]
+  where getCtorName :: Type -> Name
+        getCtorName c = case c of
+          ConT nm -> nm
+          _ -> error $ "mkKnownReprs cannot be used on polymorphic data kinds."
 
 mkReprName :: Name -> Name
 mkReprName nm = mkName (nameBase nm ++ "Repr")
-
-getCtorName :: Type -> Name
-getCtorName c = case c of
-  ConT nm -> nm
-  _ -> error $ "expected constructor, found " ++ show c
 
 -- $typePatterns
 --
--- a/src/Data/Parameterized/TH/GADT.hs
+++ b/src/Data/Parameterized/TH/GADT.hs
@@ -25,6 +25,10 @@ module Data.Parameterized.TH.GADT
   , structuralHash
   , structuralHashWithSalt
   , PolyEq(..)
+    -- * Repr generators ("singletons")
+    -- $reprs
+  , mkRepr
+  , mkKnownReprs
     -- * Template haskell utilities that may be useful in other contexts.
   , DataD
   , lookupDataType'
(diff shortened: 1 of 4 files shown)


src/Data/Parameterized/TH/GADT.hs
```diff
@@ -493,6 +497,151 @@ showCon p nm n = do
...
+--
+-- @
+-- data T4 = T4_1 Nat | T4_2 Peano
+-- \$(mkRepr ''T4)
+-- ======>
+-- data T4Repr (tp :: T4)
+--   where
+--     T4Repr :: NatRepr tp1 -> PeanoRepr tp2 -> T4Repr ('T4 tp1 tp2)
+-- @
+mkRepr :: Name -> DecsQ
```
RyanGlScott, line 549: I'm not sure how robust this function is with respect to polymorphism. What will it do on a data type like this, for example?

```hs
data Foo a b = MkFoo a b
```

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -532,3 +681,40 @@ matchShowCtor p con = showCon p (constructorName con) (length (constructorFields
...
+-- we have a corresponding
+--
+-- @
+-- data TRepr (t :: T) where
+--   ARepr :: TRepr 'A
+--   BRepr :: NatRepr w -> TRepr ('B w)
+-- @
+--
+-- Assuming the user of @parameterized-utils@ follows this convention, we
+-- provide the template haskell constructs 'mkRepr' to automate the creation of
```
RyanGlScott, line 717:
```suggestion
-- provide the Template Haskell construct 'mkRepr' to automate the creation of
```

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -532,3 +681,40 @@ matchShowCtor p con = showCon p (constructorName con) (length (constructorFields
...
+-- the @TRepr@ GADT type. We also provide 'mkKnownReprs', which generates
+-- 'KnownRepr' instances for that GADT type. See the documentation for those two
```
RyanGlScott, line 719: "GADT type" is a bit like "ATM machine" in that the last word is redundant :)

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -493,6 +497,151 @@ showCon p nm n = do
...
+            ctorType = AppT (ConT reprTypeName) appliedType
+            ctorArgTypes =
+              zipWith (\n v -> (Bang NoSourceUnpackedness NoSourceStrictness, AppT (ConT n) (VarT v))) ctorFieldReprNames tvars
+        return $ GadtC
+          [reprCtorName]
+          ctorArgTypes
+          ctorType
+  ctors <- mapM gc (datatypeCons info)
+  return $ [ DataD [] reprTypeName
+             [KindedTV varName () (ConT typeName)]
```
RyanGlScott, line 575: Before GHC 9.0, `KindedTV` didn't have `()` as its second field. To make this backwards-compatible with earlier versions of GHC, I'd recommend using [`kindedTV`](https://hackage.haskell.org/package/template-haskell-2.17.0.0/docs/Language-Haskell-TH-Lib.html#v:kindedTV) from `Language.Haskell.TH.Lib`, which does the right thing on all versions of GHC.

RyanGlScott (commented):
As you've noted, this is mimicking quite a bit of functionality from the `singletons-th` library. This isn't necessarily a bad thing, but perhaps the documentation ought to include a reference to `singletons-th` for the sake of completeness.

src/Data/Parameterized/TH/GADT.hs
benjaminselfridge: Good point :)

src/Data/Parameterized/TH/GADT.hs
benjaminselfridge: Yes, I noticed the 8.* failures. Thanks for identifying the problem.

src/Data/Parameterized/TH/GADT.hs
benjaminselfridge: Oh, I see now. Thanks! `parameterized-utils` does not have anything analogous to `Sing`, and I don't think we have a need for it necessarily; most of the use cases I've seen involve plain, basically monomorphic data kinds (the types you see in `crucible` and `what4` are typically as complex as it gets).

I did not think very hard about generating good error messages, so I need to put a bit more effort into that. I think we basically want a check for data kinds that makes sure they are totally monomorphic, and it might also be a good idea to make sure that all the types they are built from also have corresponding `repr` types.

RyanGlScott: I suppose an alternative approach would be to do something like what the [`List` repr does](http://hackage.haskell.org/package/parameterized-utils-2.1.3.0/docs/Data-Parameterized-List.html#t:List) and parameterize it by another type variable `f`:

```hs
data List :: (k -> Type) -> [k] -> Type where
  Nil  :: List f '[]
  (:<) :: f tp -> List f tps -> List f (tp : tps)
```

You'd need a type variable like `f` for each type parameter your original data type has.

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -493,6 +497,163 @@ showCon p nm n = do
...
+-- @
+mkKnownReprs :: Name -> DecsQ
+mkKnownReprs typeName = do
+  kr <- [t|KnownRepr|]
+  let krFName = mkName "knownRepr"
+      reprTypeName = mkReprName typeName
+  typeInfo <- lookupDataType' typeName
+  reprInfo <- lookupDataType' reprTypeName
+  forM (zip (datatypeCons typeInfo) (datatypeCons reprInfo)) $ \(tci, rci) -> do
+    vars <- forM (constructorFields tci) $ const (newName "tp")
```
RyanGlScott, line 637: This:

```suggestion
    vars <- replicateM (length (constructorFields tci)) (newName "tp")
```

Is likely slightly more efficient.

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -493,6 +497,163 @@ showCon p nm n = do
...
+        krConstraint = AppT (AppT kr (ConT reprTypeName)) appliedType
+        krExp = foldl AppE (ConE (constructorName rci)) $
+          map (const (VarE krFName)) vars
+        krDec = FunD krFName [Clause [] (NormalB krExp) []]
+
+    return $ InstanceD Nothing krReqs krConstraint [krDec]
+  where getCtorName :: Type -> Name
+        getCtorName c = case c of
+          ConT nm -> nm
+          _ -> error $ "mkKnownReprs cannot be used on polymorphic data kinds."
```
RyanGlScott, line 652: I like the error message here, although it will likely trigger on things besides type variables, such as `ListT` (e.g., `[]`) or `AppT` (e.g., `Maybe Int`). Perhaps reserver the `mkKnownReprs cannot be used on polymorphic data kinds` error for `VarT` and have a generic catch-all error for everything else.

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -25,6 +25,10 @@ module Data.Parameterized.TH.GADT
   , structuralHash
   , structuralHashWithSalt
   , PolyEq(..)
+    -- * Repr generators ("singletons")
```
RyanGlScott, line 28: Haddock will treat anything between double quotes as a module name, which isn't what you want here. Do this instead:

```suggestion
    -- * Repr generators (\"singletons\")
```

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -493,6 +497,163 @@ showCon p nm n = do
 matchShowCtor :: ExpQ -> ConstructorInfo -> MatchQ
 matchShowCtor p con = showCon p (constructorName con) (length (constructorFields con))
 
+-- | Generate a "repr" or singleton type from a data kind. For nullary
```
RyanGlScott, line 500:
```suggestion
-- | Generate a \"repr\" or singleton type from a data kind. For nullary
```

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -493,6 +497,163 @@ showCon p nm n = do
...
+-- @
+-- data T3 = T3 T1 T2
+-- \$(mkRepr ''T3)
+-- ======>
+-- data T3Repr (tp :: T3)
+--   where
+--     T3Repr :: T1Repr tp1 -> T2Repr tp2 -> T3Repr ('T3 tp1 tp2)
+-- @
+--
+-- This is generally compatible with other "repr" types provided by
```
RyanGlScott, line 538:
```suggestion
-- This is generally compatible with other \"repr\" types provided by
```

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -493,6 +497,163 @@ showCon p nm n = do
...
+--   knownRepr = T2_1Repr knownRepr
+-- @
+--
+-- @
+-- \$(mkKnownReprs ''T3)
+-- ======>
+-- instance (KnownRepr T1Repr tp1, KnownRepr T2Repr tp2) =>
+--          KnownRepr T3Repr ('T3_1 tp1 tp2) where
+--   knownRepr = T3_1Repr knownRepr knownRepr
+-- @
```
RyanGlScott, line 628: The same restrictions for `mkRepr` also apply to `mkKnownReprs`, right? If so, mention them here in the Haddocks for `mkKnownReprs`.

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -532,3 +693,44 @@ matchShowCtor p con = showCon p (constructorName con) (length (constructorFields
...
+--
+-- Assuming the user of @parameterized-utils@ follows this convention, we
+-- provide the Template Haskell construct 'mkRepr' to automate the creation of
+-- the @TRepr@ GADT. We also provide 'mkKnownReprs', which generates 'KnownRepr'
+-- instances for that GADT type. See the documentation for those two functions
+-- for more detailed explanations.
+--
+-- NB: These macros are inspired by the corresponding macros provided by
+-- `singletons-th`, and the "repr" programming idiom is very similar to the one
```
RyanGlScott, line 735:
```suggestion
-- @singletons-th@, and the \"repr\" programming idiom is very similar to the one
```

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -532,3 +693,44 @@ matchShowCtor p con = showCon p (constructorName con) (length (constructorFields
...
+-- used by `singletons.`
```
RyanGlScott, line 736:
```suggestion
-- used by @singletons@.
```

src/Data/Parameterized/TH/GADT.hs
benjaminselfridge: Great, thanks

RyanGlScott (approved):
LGTM.

PR #115, opened 2021-05-19
Factor out `unsafeCoerce Refl` into its own definition

### LINKED ISSUE #114: `lemmaMul` is unsafe
The definition of `lemmaMul`:

https://github.com/GaloisInc/parameterized-utils/blob/13f367bc029de35f9ab4edec2bcd90bc638d30a7/src/Data/Parameterized/NatRepr.hs#L615-L617

is unsafe, as it unsafely coerces `Refl` from type `a :~: a` to the function type `p w -> q n -> (w + (n-1) * w) :~: (n * w)`. In practice, this leads to a runtime crash when `parameterized-utils` is compiled without optimization:

```
> import qualified Data.Parameterized.Vector as V                                              
import qualified Data.Parameterized.Vector as V                                                
> import Data.Parameterized.NatRepr                                                            
import Data.Parameterized.NatRepr                                                              
> Just v = V.fromList (knownNat @2) [knownNat @1, knownNat @1]                                 
Just v = V.fromList (knownNat @2) [knownNat @1, knownNat @1]                                   
                                                                                               
<interactive>:3:1: warning: [-Wincomplete-uni-patterns]                                        
    Pattern match(es) are non-exhaustive                                                       
    In a pattern binding: Patterns not matched: Nothing                                        
> V.joinWith (\_ w l -> addNat w l) knownNat v                                                 
V.joinWith (\_ w l -> addNat w l) knownNat v                                                   
<interactive>: internal error: stg_ap_pp_ret                                                   
    (GHC version 8.8.4 for x86_64_apple_darwin)                                                
    Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug                   
cabal: repl failed for parameterized-utils-2.1.3.0.100. The build process                      
terminated with exit code -6
```

The solution is simple: add two arguments to the definition of `lemmaMul`. Patch incoming.

_ht to the good folks at the `#ghc` IRC channel for helping to spot this bug._

RyanGlScott:
Fixes #114.

### DIFF
--- a/parameterized-utils.cabal
+++ b/parameterized-utils.cabal
@@ -69,6 +69,7 @@ library
   exposed-modules:
     Data.Parameterized
     Data.Parameterized.All
+    Data.Parameterized.Axiom
     Data.Parameterized.BoolRepr
     Data.Parameterized.Classes
     Data.Parameterized.ClassesC
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,16 @@
 # Changelog for the `parameterized-utils` package
 
+## next -- *unreleased*
+
+  * Fix a bug in which `Data.Parameterized.Vector.{join,joinWith,joinWithM}`
+    and `Data.Parameterized.NatRepr.plusAssoc` could crash at runtime if
+    compiled without optimizations.
+  * Add a `Data.Parameterized.Axiom` module providing `unsafeAxiom` and
+    `unsafeHeteroAxiom`, which can construct proofs of equality between types
+    that GHC isn't able to prove on its own. These functions are unsafe if used
+    improperly, so the responsibility is on the programmer to ensure that these
+    functions are used appropriately.
+
 ## 2.1.3.0 -- *2021 Mar 23*
 
   * Add support for GHC 9.
--- a/src/Data/Parameterized/Axiom.hs
+++ b/src/Data/Parameterized/Axiom.hs
@@ -1,0 +1,48 @@
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE Unsafe #-}
+{-|
+Copyright        : (c) Galois, Inc 2014-2021
+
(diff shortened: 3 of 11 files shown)


RyanGlScott:
There's yet another example of this same sort of issue in `plusAssoc`:

https://github.com/GaloisInc/parameterized-utils/blob/b2e9cc34a58e5931cf14beca2c5bc28f343cf966/src/Data/Parameterized/NatRepr.hs#L307-L309

I should do a more thorough audit of where `unsafeCoerce Refl` is used. Until I do, I'll mark this as a draft.

RyanGlScott:
I went the extra mile and factored out `unsafeCoerce Refl` into its own definition, `axiom`, in the new `Data.Parameterized.Axiom` module. I also replaced every occurrence of `unsafeCoerce Refl` with `axiom` that I could find, fixing the bug observed in https://github.com/GaloisInc/parameterized-utils/pull/115#issuecomment-844321035 in the process. We could go further and define variants of `axiom` for other proof data types (e.g., `LeqProof`), but I'll leave that as future work.

This is now ready for review.

parameterized-utils.cabal
```diff
@@ -69,6 +69,7 @@ library
   exposed-modules:
     Data.Parameterized
     Data.Parameterized.All
+    Data.Parameterized.Axiom
```
RyanGlScott, line 72: I went ahead and exported this, since it's conceivable that a user might want to use `axiom` to construct proofs that aren't provided by `parameterized-utils`. I don't have a strong opinion on this, however, and I would be find with making it an internal module if others feel differently.

travitch: Thanks for doing this! I've wanted something like `axiom` before just to make it clear what was going on, so I like exporting it to provide a nice common name. It seems to be guarded appropriately by scary comments.

simonjwinwood: Would unsafeAxiom be a better name?

RyanGlScott: Sounds good to me. I've pushed a change renaming `axiom`/`haxiom` to `unsafeAxiom`/`unsafeHeteroAxiom`.

langston-barrett (approved):

PR #117, opened 2021-09-14
TraversableFCWithIndex: Indexed traversals, folds, and maps

langston-barrett:
Add indexed functors, folds, and traversals inspired by the indexed-traversable
(which is where the naming scheme comes from) and lens packages.

The current contents of the List and Assignment modules provide additional
motivation: List already implements and exports a minimal set of methods for
each of the added classes, but Assignment only implemented
'traverseWithIndex' (AKA 'itraverseFC'); this indicates that identifying this
abstraction enables code sharing and increased functionality between these
similar modules/datatypes.

Additionally, Assignment now gets 'ifoldMapFC', 'imapFC' and friends "for free",
because they have default implementations based on 'itraverseFC'.

At some point, it may make sense to implement similar classes for the *F family:
'FunctorF', 'FoldableF', and 'TraversableF'. However, 'MapF' is the only
'TraversableF' that has an instance of 'IndexF', so this isn't as clearly a
code-sharing win, though such classes may uncover useful, as of yet
unimplemented functions for that API as the *FC ones did for 'Assignment'.

### DIFF
--- a/src/Data/Parameterized/Context/Safe.hs
+++ b/src/Data/Parameterized/Context/Safe.hs
@@ -124,6 +124,7 @@ import Data.Parameterized.Ctx
 import Data.Parameterized.NatRepr
 import Data.Parameterized.Some
 import Data.Parameterized.TraversableFC
+import Data.Parameterized.TraversableFC.WithIndex
 
 ------------------------------------------------------------------------
 -- Size
@@ -615,6 +616,15 @@ instance FoldableFC Assignment where
 instance TraversableFC Assignment where
   traverseFC f = traverseF f
 
+instance FunctorFCWithIndex Assignment where
+  imapFC = imapFCDefault
+
+instance FoldableFCWithIndex Assignment where
+  ifoldMapFC = ifoldMapFCDefault
+
+instance TraversableFCWithIndex Assignment where
+  itraverseFC = traverseWithIndex
+
 -- | Map assignment
 map :: (forall tp . f tp -> g tp) -> Assignment f c -> Assignment g c
 map f = fmapFC f
@@ -650,6 +660,7 @@ zipWith :: (forall x . f x -> g x -> h x)
 zipWith f = \x y -> runIdentity $ zipWithM (\u v -> pure (f u v)) x y
 {-# INLINE zipWith #-}
 
+-- | This is a specialization of 'itraverseFC'.
 traverseWithIndex :: Applicative m
                   => (forall tp . Index ctx tp -> f tp -> m (g tp))
                   -> Assignment f ctx
--- a/src/Data/Parameterized/TraversableFC/WithIndex.hs
+++ b/src/Data/Parameterized/TraversableFC/WithIndex.hs
@@ -1,0 +1,175 @@
+------------------------------------------------------------------------
+-- |
+-- Module           : Data.Parameterized.TraversableFC.WithIndex
+-- Copyright        : (c) Galois, Inc 2021
+-- Maintainer       : Langston Barrett
+-- Description      : 'TraversableFC' classes, but with indices.
+--
+-- As in the package indexed-traversable.
+------------------------------------------------------------------------
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+
+module Data.Parameterized.TraversableFC.WithIndex
+  ( FunctorFCWithIndex(..)
+  , FoldableFCWithIndex(..)
+  , ifoldlMFC
+  , ifoldrMFC
+  , iallFC
+  , ianyFC
+  , TraversableFCWithIndex(..)
+  , imapFCDefault
+  , ifoldMapFCDefault
+  ) where
+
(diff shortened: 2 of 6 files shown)


src/Data/Parameterized/TraversableFC/WithIndex.hs
```diff
@@ -0,0 +1,181 @@
...
+  Monad m =>
+  (forall x. IndexF (t f z) x -> f x -> b -> m b) ->
+  b ->
+  t f z ->
+  m b
+ifoldrMFC f z0 xs = ifoldlFC (\i k x z -> f i x z >>= k) return xs z0
+
+-- | This is a coercion used to avoid overhead associated
+-- with function composition.
+(#.) :: Coercible b c => (b -> c) -> (a -> b) -> (a -> c)
```
RyanGlScott, line 131: Since `parameterized-utils` already transitively depends on `profunctors` (by way of `lens`), you could alternatively import [`(#.)` from `Data.Profunctor.Unsafe`](https://hackage.haskell.org/package/profunctors-5.6.2/docs/Data-Profunctor-Unsafe.html#v:-35-.).

src/Data/Parameterized/TraversableFC/WithIndex.hs
```diff
@@ -0,0 +1,181 @@
...
+    t f z ->
+    m (t g z)
+
+imapFCDefault ::
+  forall t f g z.
+  TraversableFCWithIndex t =>
+  (forall x. IndexF (t f z) x -> f x -> g x)
+  -> t f z
+  -> t g z
+imapFCDefault f = runIdentity . itraverseFC (\i x -> Identity (f i x))
```
RyanGlScott, line 170: Speaking of which, why not use `(#.)` here instead of `(.)`? (Similarly for `ifoldMapFCDefault`.)

src/Data/Parameterized/Context/Safe.hs
```diff
@@ -655,6 +665,7 @@ traverseWithIndex :: Applicative m
                   -> Assignment f ctx
                   -> m (Assignment g ctx)
 traverseWithIndex f a = generateM (size a) $ \i -> f i (a ! i)
+{-# DEPRECATED traverseWithIndex "Use itraverseFC" #-}
```
RyanGlScott, line 668: I'm perhaps not completely up-to-date on the conventions employed in `parameterized-utils`, but does it make sense to all-out deprecate functions like `traverseWithIndex`? Other parts of `parameterized-utils` happily define specialized versions of class methods (e.g., [`Data.Parameterized.Map.map`](https://hackage.haskell.org/package/parameterized-utils-2.1.3.0/docs/Data-Parameterized-Map.html#v:map)).

langston-barrett: Honestly, I have no particular opinion on this - My instinct was that it made sense to have just one name for these things, but I'd be happy to keep around the specializations.

langston-barrett:
I'm not sure what's going on with the CI here:

> cabal: Cannot test the package parameterized-utils-2.1.3.0.100 because none of
> the components are available to build: the test suite 'parameterizedTests' is
> not available because the solver did not find a plan that included the test
> suites. Force the solver to enable this for all packages by adding the line
> 'tests: True' to the 'cabal.project.local' file.

src/Data/Parameterized/TraversableFC/WithIndex.hs
langston-barrett: Thanks, I wasn't aware of that function!

src/Data/Parameterized/TraversableFC/WithIndex.hs
langston-barrett: Good point, done!

RyanGlScott:
This isn't your fault, as it's also happening on `master` as well. This is a odd feature of `cabal-install` where by default, it will sometimes default to build plans that exclude test suite dependencies if they result in a build plan that is somehow deemed less optimal than a build plan that only includes the library dependencies. (I'm far from an expert on `cabal-install`'s internals, so I have no idea what algorithm determines the optimality of a build plan.)

In any case, the fix is to use `cabal`'s `--enable-tests` flag, which forces the solver to consider test suite dependencies. Most CI configurations for our various repos run `cabal configure --enable-tests` up front to make sure this happens, so I'm surprised that `parameterized-utils` doesn't do the same.

langston-barrett:
If #119 looks good, then I can rebase this PR onto those changes after merging them and we can get an accurate read from CI before proceeding.

src/Data/Parameterized/Context/Safe.hs
RyanGlScott: Considering that `traverseWithIndex` is fairly commonly used various projects (see [here](https://github.com/search?q=%22traverseWithIndex%22+user%3AGaloisInc&type=code)), my inclination is to avoid the code churn that a deprecation cycle would induce.

langston-barrett: I replaced the deprecations with notes in the Haddocks. Thanks for the note, I really hadn't considered the impact on downstream projects.

kquick (approved):
These are strictly enhancements and so I don't anticipate any downstream negative impact.

PR #120, opened 2021-09-16
Document FunctorFC laws, test Assignment instance

langston-barrett:
I realized that the laws and tests I wrote for #117 implicitly relied on the assumption that the other *FC classes had laws and tests for those laws. This isn't all of them, but it's a start.

Also, strengthen a few existing *WithIndex tests.


kquick (approved):
Looks good!

PR #122, opened 2021-09-24
Fin: Finite types

langston-barrett:
The type 'VectorIndex' was named something more specific than it needed to be. Another way to look at it is that it's the definition of the family of finite types.

### DIFF
--- a/src/Data/Parameterized/Fin.hs
+++ b/src/Data/Parameterized/Fin.hs
@@ -1,0 +1,111 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE EmptyCase #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+
+{-|
+Copyright        : (c) Galois, Inc 2021
+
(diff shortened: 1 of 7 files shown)


src/Data/Parameterized/Fin.hs
```diff
@@ -0,0 +1,84 @@
...
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+
+{-|
+Copyright        : (c) Galois, Inc 2014-2019
+
+Finite types.
```
kquick, line 13: Could you provide more of an explanation and/or some examples here?  This becomes a sufficiently abstract representation that the utility is not immediately obvious.  It might be nice to add something to the README for p-u as well describing this.

Also, I think the copyright date on this particular module should be "2021".

kquick (approved):
Looks fine in general, just a request for a bit more documentation.

src/Data/Parameterized/Fin.hs
langston-barrett: Yes, will do! In short, my motivation for introducing these is to use them as indices into `Vector`.

langston-barrett: I added more detail here as well as a blurb in the README, let me know what you think.

kquick: That's helpful, thanks!

PR #128, opened 2022-01-31
Support building with GHC 9.2

RyanGlScott:
This contains a variety of changes needed to make `parameterized-utils` compile
with GHC 9.2:

* In `base-4.16.*`, the internals of the `(<=)` type family in `GHC.TypeNats` have changed (see [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.2?version_id=7e2ce63ba042c1934654c4316dc02028d8d3dd31#changes-to-the-and-type-families)). As a result, GHC is no longer able to infer `0 <= n` for all `n :: Nat`, as this fact must now be assumed as an axiom. This prevents a handful of functions in `Data.Parameterized.NatRepr` and `Data.Parameterized.Vector` from compiling, as they relied on `0 <= n` to typecheck.

  I resolved this issue by introducing a new `leqZero :: LeqProof 0 n` function to `Data.Parameterized.NatRepr` and using it where needed.
* In `template-haskell-2.18.*`, the type of `ConP` gained an additional field (see [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.2?version_id=7e2ce63ba042c1934654c4316dc02028d8d3dd31#template-haskell-218)). As a result, I needed to use some CPP in `Data.Parameterized.TH.GADT` to make the two uses of `ConP` compile. To minimize the amount of CPP that I needed, I factored out this logic into a `conPCompat` function.
* In GHC 9.2, enabling `UndecidableInstances` no longer implies enabling `FlexibleContexts` (see [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.2?version_id=7e2ce63ba042c1934654c4316dc02028d8d3dd31#undecidableinstances-no-longer-implies-flexiblecontexts-in-instance-declarations)). As a result, I had to enable `FlexibleContexts` in `Data.Parameterized.Context.Unsafe`.
* The upper version bounds on `lens` have been raised to allow building with `lens-5.1`, which is the minimum version of `lens` that builds with GHC 9.2.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        ghc: [9.2.1, 9.0.2, 8.10.7, 8.8.4, 8.6.5, 8.4.4]
+        ghc: [9.2.1, 9.0.2, 8.10.7, 8.8.4, 8.6.5]
         allow-failure: [false]
       fail-fast: false
     steps:
@@ -47,7 +47,6 @@ jobs:
           GHC=haskell.compiler.ghc$(echo ${{ matrix.ghc }} | sed -e s,\\.,,g)
           echo GHC=$GHC >> $GITHUB_ENV
           case ${{ matrix.ghc }} in
-            8.4.4) GHC_NIXPKGS=github:nixos/nixpkgs/20.03 ;;
             8.6.5) GHC_NIXPKGS=github:nixos/nixpkgs/20.09 ;;
             9.0.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-unstable ;;
             9.2.1) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-unstable ;;
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        ghc: [9.0.1, 8.10.4, 8.8.4, 8.6.5, 8.4.4]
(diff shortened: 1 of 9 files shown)


kquick (approved):
Looks good.

Did you drop GHC 8.4 support because it's no longer compatible or just as an update to the trailing support horizon?

kquick (approved):
Looks good.

Did you drop GHC 8.4 support because it's no longer compatible or just as an update to the trailing support horizon?

RyanGlScott:
> Did you drop GHC 8.4 support because it's no longer compatible or just as an update to the trailing support horizon?

Mainly because of issues trying to get Nix flakes support for 8.4. I chronicled my woes in this commit message:

> **CI: Stop testing GHC 8.4**
>
> Unfortunately, it's quite hard to test GHC 8.4 with a Nix Flake–based setup.
Attempting to do so with NixOS 20.03, a version of NixOS that includes GHC
8.4.4, results in the following error:
>
> ```
> error: flake 'github:nixos/nixpkgs/5272327b81ed355bbed5659b8d303cf2979b6953' has an unsupported attribute 'edition', at /nix/store/1q41mjncrlsds9nxwr2xrl6lb9g48man-source/flake.nix:4:3
> ```
>
> This means that this version of NixOS is likely too old. Since newer versions
of NixOS don't include GHC 8.4.4, let's just make things easier and not include
GHC 8.4 in CI.

If you have ideas on how this might be possible, I'm all ears. Then again, I'm not sure if it's worth the trouble.

kquick:
Ahh, I'd missed the transition from pre-flakes nix (`nix-shell`) to flake-supporting nix (`nix shell`), and yes, 20.03 quite likely predates the ability to support flakes.

Sorry I missed the commit message info, and--while we could potentially support 8.4 via use of `nix-shell` for that version--I'd agree that it's not worth the trouble.  Please feel free to merge when you're ready!

PR #131, opened 2022-04-11
Replace `testProperty` with `testPropertyNamed`

RyanGlScott:
`tasty-hedgehog-1.2.0.0` has deprecated the [`testProperty`](https://hackage.haskell.org/package/tasty-hedgehog-1.2.0.0/docs/Test-Tasty-Hedgehog.html#v:testProperty) function in favor of [`testPropertyNamed`](https://hackage.haskell.org/package/tasty-hedgehog-1.2.0.0/docs/Test-Tasty-Hedgehog.html#v:testPropertyNamed), which expects the name of the `Property` value being tested as an argument. To avoid heaps of deprecation warnings when building the test suite, let's switch from `testProperty` to `testPropertyNamed`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        ghc: [9.2.1, 9.0.2, 8.10.7, 8.8.4, 8.6.5]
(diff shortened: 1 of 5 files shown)


kquick:
That was a lot of annoying mechanical changes; thanks for taking care of this!

kquick (approved):

Issue #129, opened 2022-02-01
TestEquality clarification from core library devs


robdockins:
The following patch has been approved and should eventually make it's way into `base`.

https://gitlab.haskell.org/ghc/ghc/-/merge_requests/7333/diffs

It clarifies that `testEquality` should return `Just Refl` _iff_ the corresponding type indices are the same, and reiterates that it is desirable that the class only be inhabited by singletons.  This is incompatible with many of our typical uses of this class.
 
The actual class declaration will not change, so nothing will immediately break with future releases of base. However, going forward we will have to decide what to do about the fact that many of our uses of `TestEquality` will be considered law-breaking.

travitch:
Fun. Since nearly all of our uses import the required class via `Data.Parameterized.Classes` (instead of the type equality module in base), we could always just duplicate the class definition with dire warnings that they are not the same. That would save us from having to change much client code.

That might ultimately be bad and confusing, though.

robdockins:
I think duplicating the class and reusing the name would be strictly worse than doing nothing and having law-breaking instances.

benjaminselfridge:
If I could go back in time, I would change `EqF` to correspond with `OrdF`. Our `EqF` is (I think) not really used much, because you have to know that the type indices are the same at compile time, whereas `OrdF` is heterogeneous, and the result gives you information both at the type- and value-level. `EqF` seems like a natural name for the corresponding equality type class.

Of course, going back and auditing all uses of `TestEquality` in Galois's codebase would be a nightmare. But if I had magical powers, that's probably what I would do.

I am not a huge fan of making our own `TestEquality` that conflicts with base, because I think `TestEquality` is already a bad name for the way we use it, and if we're going to fix the problem we should put our big boy pants on and fix the name while we're at it.

robdockins:
I agree, `EqF` isn't terribly useful as-is. I'd be in favor of repurposing it, unless it is widely used in some way I'm not aware of.

Fortunately, we can make this transition gradually if we decide to do it, since `TestEquality` isn't going anywhere and we can have side-by-side instances for a time.

benjaminselfridge:
I think repurposing `EqF` for this seems like a good idea; the only concern I have is that I'm not sure if `EqF` is actually used anywhere. But I'd be willing to spend a few hours investigating this.

robdockins:
[...]

PR #132, opened 2022-04-19
Draft: Redefines the EqF type class to be heterogeneous.

benjaminselfridge:
This is just a draft -- not ready for merge.

This implements a change discussed in #129. Basically, it redefines `EqF` to be same-kinded with `TestEquality`, since the homogeneous variant wasn't being used. The idea is that `EqF` asks "are the types AND values of these two things equal?" so it makes sense to define on non-singletons.

I tried to be relatively conservative here, keeping `TestEquality` instances in most cases.

### DIFF
--- a/src/Data/Parameterized/Classes.hs
+++ b/src/Data/Parameterized/Classes.hs
@@ -25,7 +25,7 @@ not restricted to '*'.
 {-# LANGUAGE TypeOperators #-}
 module Data.Parameterized.Classes
   ( -- * Equality exports
-    Equality.TestEquality(..)
+    TestEquality(..)
   , (Equality.:~:)(..)
   , EqF(..)
   , PolyEq(..)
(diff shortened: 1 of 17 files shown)


src/Data/Parameterized/List.hs
```diff
@@ -267,10 +275,10 @@ deriving instance Show  (Index l x)
 
 instance ShowF (Index l)
 
-instance TestEquality (Index l) where
-  testEquality IndexHere IndexHere = Just Refl
-  testEquality (IndexThere x) (IndexThere y) = testEquality x y
-  testEquality _ _ = Nothing
```
RyanGlScott, line 273: I presume that this instance was removed because it does not obey the `TestEquality` laws? If so, should we remove the instances for other `Index` data types in `parameterized-utils` (e.g., `Data.Parameterized.Context.{Safe,Unsafe}.Index`)? Or would it be better to keep this `TestEquality` instance for a while to give users a chance to migrate to `EqF`?

src/Data/Parameterized/Classes.hs
```diff
@@ -90,27 +90,23 @@ instance CoercibleF (Const x) where
...
---
--- Unlike 'TestEquality', this only works when the type arguments are
--- the same, and does not provide a proof that the types have the same
--- type when they are equal. Thus this can be implemented over
--- parameterized types that are unable to provide evidence that their
--- type arguments are equal.
+-- | @EqF@ allows you to test for type- and value-level equality over a
+-- parameterized type. It is like @TestEquality@, but unlike that class, @EqF@
+-- returns `Nothing` when the type indices are equal but the values are not.
+-- This makes it appropriate to implement on non-singleton types.
```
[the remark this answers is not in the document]
benjaminselfridge: Good call.

src/Data/Parameterized/List.hs
benjaminselfridge: Yeah, and I actually meant to put this back in to allow just such a gradual change.

Initially I had actually removed the instances in `Context`, but realized that it might screw people up in the short term if I did that.

Do you know if there is a way to deprecate a type class instance and have GHC give a warning that they should use `eqF` instead of `testEquality`?

benjaminselfridge: Yep, this would entail a lot of churn -- not only will we uncover missing `EqF` instances downstream, this will invalidate all existing `EqF` instances (although I'm not sure if there are too many of those).

But yeah, the solution should just be to copy-paste instances.

I made this PR in discussion with RobD, as a way to start the conversation about whether this change is worth it. I think it is worth it, personally -- insofar as `parameterized-utils` is Galois' offering of a "standard library for dependent types in Haskell", we want to be in compliance with all the basic laws.

robdockins:
I'm generally in favor of this change.  However, I think we need to break it into pieces to help manage the churn.  First, let's change the Kind of `EqF`, implement new instances here and give downstream packages some time to implement their own instances.  Next, we can change things that currently require `TestEquality` to require `EqF` instead, and finally we can remove `TestEquality` instances that don't follow the laws.

robdockins:
I'm generally in favor of this change.  However, I think we need to break it into pieces to help manage the churn.  First, let's change the Kind of `EqF`, implement new instances here and give downstream packages some time to implement their own instances.  Next, we can change things that currently require `TestEquality` to require `EqF` instead, and finally we can remove `TestEquality` instances that don't follow the laws.

benjaminselfridge:
Makes sense to me.

RyanGlScott:
I have opened #174 separately to introduce the `EqF` instances that this PR adds, but without making any changes to the definition of the `EqF` class itself.

PR #133, opened 2022-04-20
FinMap: an integer map with a statically-known maximum size

langston-barrett:
This PR introduces a new interface for a type `FinMap n a`, which is conceptually a map with keys in `Fin n`, i.e., a map with a maximum statically known key/size.

My specific motivation to create these modules is from UC-Crux. Currently, [UC-Crux inefficiently](https://github.com/GaloisInc/crucible/blob/041975a5ad4200633613ed30df144c3a9d0e9403/uc-crux-llvm/src/UCCrux/LLVM/Shape.hs#L120) tracks preconditions for array-typed values in a `Vector n whatever`. However, many of the `whatever` values are effectively "defaults", so this representation is space-inefficient.

This is still a draft (as I would want to write more tests and documentation before merging), but I'd love your thoughts as to the overall structure/approach @kquick. In particular,

- Do we like the structure of having a safe and unsafe implementation, as is done for `Data.Parameterized.Context`? On the one hand, it's somewhat wasteful to spend time developing and compiling two implementations of the same functionality. On the other hand, it ensures any operations that internally use unsafe functions could feasibly be implemented without them (demonstrating that they are at least potentially sound), and it provides a strong basis for property-based testing.
- Are these the right choices of representation/concrete datatype? `IntMap` is nice because it's fast, but annoying because of its boundedness. A few other possible choices:
  *  We could represent both safe and unsafe versions as `Map (Fin n)`, but have the unsafe version use unsafe operations where the safe version provides proofs
  * We could represent the unsafe version with a `Map Integer`
- The bit about `Int` overflow is annoying... I didn't add checks to prevent it (for the sake of performance), but I'm also not entirely satisfied with the approach of assuming it doesn't happen. Do you have thoughts on how best to approach this?
- What about the name? Maybe `FinMap` would be better...

### DIFF
--- a/src/Data/Parameterized/FinMap.hs
+++ b/src/Data/Parameterized/FinMap.hs
@@ -1,24 +1,24 @@
 {-|
 Copyright        : (c) Galois, Inc 2022
 
-'SizedMap' is a map with 'NatRepr' keys and a maximum size reflected in its
-type. @'SizedMap' n a@ can be used as a more space-efficient replacement for a
+'FinMap' is a map with 'NatRepr' keys and a maximum size reflected in its
+type. @'FinMap' n a@ can be used as a more space-efficient replacement for a
 @'Data.Parameterized.Vector.Vector' n ('Maybe' a)@, or a replacement for an
 @'Data.IntMap.IntMap' a@ where the maximum key (i.e., size) of the map can be
 tracked statically.
 
 This interface has two implementations:
 
-* 'Data.Parameterized.SizedMap.Unsafe.SizedMap' is backed by an
+* 'Data.Parameterized.FinMap.Unsafe.FinMap' is backed by an
   'Data.IntMap.IntMap', and must have a size of at most @'maxBound' :: 'Int'@.
   This module uses unsafe operations like 'Unsafe.Coerce.unsafeCoerce'
   internally for the sake of time and space efficiency.
-* 'Data.Parameterized.SizedMap.Safe.SizedMap' is backed by an
+* 'Data.Parameterized.FinMap.Safe.FinMap' is backed by an
   @'Data.Map.Map' ('Data.Parameterized.Fin.Fin' n)@. All of its functions are
   implemented using safe operations.
 
-The implementation in 'Data.Parameterized.SizedMap.Unsafe.SizedMap' is property
-tested against that in 'Data.Parameterized.SizedMap.Safe.SizedMap' to ensure
+The implementation in 'Data.Parameterized.FinMap.Unsafe.FinMap' is property
+tested against that in 'Data.Parameterized.FinMap.Safe.FinMap' to ensure
 they have the same behavior.
 
 /W/ is used in big-O notations the same way as in the "Data.IntMap"
@@ -34,17 +34,17 @@ documentation.
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeOperators #-}
 
-module Data.Parameterized.SizedMap
+module Data.Parameterized.FinMap
   (
 #ifdef UNSAFE_OPS
-    module Data.Parameterized.SizedMap.Unsafe
+    module Data.Parameterized.FinMap.Unsafe
 #else
-    module Data.Parameterized.SizedMap.Safe
+    module Data.Parameterized.FinMap.Safe
 #endif
   ) where
 
 #ifdef UNSAFE_OPS
-import Data.Parameterized.SizedMap.Unsafe
+import Data.Parameterized.FinMap.Unsafe
 #else
-import Data.Parameterized.SizedMap.Safe
+import Data.Parameterized.FinMap.Safe
 #endif
--- a/src/Data/Parameterized/FinMap.hs
+++ b/src/Data/Parameterized/FinMap.hs
@@ -37,6 +37,16 @@ In this documentation, /W/ is used in big-O notations the same way as in the
 NOTE: Where the word "conceptually" is used, it implies that this correspondence
 is not literally true, but is true modulo some details such as differences
 between bounded types like 'Int' and unbounded types like 'Integer'.
+
+Several of the functions in both implementations are marked @INLINE@ or
+@INLINABLE@. There are three reasons for this:
+
+* Some of these just have very small definitions and so inclining is likely more
+  beneficial than harmful.
+* Some participate in @RULES@ relevant to functions used in their
+  implementations.
+* They are thin wrappers (often just newtype wrappers) around functions marked
+  @INLINE@, which should therefore also be inlined.
 -}
 
 {-# LANGUAGE CPP #-}
--- a/src/Data/Parameterized/FinMap.hs
+++ b/src/Data/Parameterized/FinMap.hs
@@ -7,15 +7,20 @@ map-like types:
 
 * @'FinMap' n a@ is conceptually isomorphic to a
   @'Data.Parameterized.Vector' n ('Maybe' a)@, but can be more space-efficient
-  especially if @n@ is large and the vector is sparsely populated with 'Just'
-  values.
+  especially if @n@ is large and the vector is populated with a small number of
+  'Just' values.
 * @'FinMap'@ is less general than 'Data.Map.Map', because it has a fixed key
   type (@'Data.Parameterized.Fin.Fin' n@).
 * @'FinMap' n a@ is similar to @'Data.IntMap.IntMap' a@, but it provides a
   static guarantee of a maximum size, and its operations (such as 'size') allow
   the recovery of more type-level information.
 * @'FinMap'@ is dissimilar from "Data.Parameterized.Map.MapF" in that neither
-  the key nor value type is parameterized.
+  the key nor value type of 'FinMap' is parameterized.
+
+The type parameter @n@ doesn't track the /current/ number of key-value pairs in
+a @'FinMap' n@ (i.e., the size of the map), but rather /an upper bound/.
+'insert' and 'delete' don't alter @n@, whereas 'incMax' does - despite the fact
+that it has no effect on the keys and values in the 'FinMap'.
 
 The 'FinMap' interface has two implementations:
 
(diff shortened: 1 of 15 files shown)


langston-barrett:
@kquick This is ready for another round of review!

src/Data/Parameterized/FinMap/Safe.hs
```diff
@@ -0,0 +1,209 @@
...
+
+-- | /O(log n)/. Fetch the value at the given key in the map.
+lookup :: Fin n -> FinMap n a -> Maybe a
+lookup k = Map.lookup k . getFinMap
+{-# INLINABLE lookup #-}
+
+-- | /O(nlog(n))/. Number of elements in the map.
+size :: forall n a. FinMap n a -> Fin (n + 1)
+size fm =
+  Fin.countFin (maxSize fm) (\k _count -> isJust (lookup (Fin.mkFin k) fm))
```
kquick, line 106: Could this be `O(n)` if implemented as  `length $ filter isJust (snd <$> Map.elems (getFinMap fm))`, assuming a `Fin n` could be directly constructed from the `Int` obtained?

kquick, line 106: Could this be `O(n)` if implemented as  `length $ filter isJust (snd <$> Map.elems (getFinMap fm))`, assuming a `Fin n` could be directly constructed from the `Int` obtained?

src/Data/Parameterized/FinMap/Unsafe.hs
```diff
@@ -0,0 +1,218 @@
...
+{-# INLINE empty #-}
+
+-- | /O(1)/. A map with one element.
+singleton :: a -> FinMap 1 a
+singleton = FinMap . IntMap.singleton 0
+{-# INLINABLE singleton #-}
+
+-- | /O(min(n,W))/.
+insert :: Fin n -> a -> FinMap n a -> FinMap n a
+insert k v = FinMap . IntMap.insert (unsafeFinToInt k) v . getFinMap
```
kquick, line 168: `insert k = fmap (IntMap.insert (unsafeFinToInt k))`  ?

kquick, line 168: `insert k = fmap (IntMap.insert (unsafeFinToInt k))`  ?

src/Data/Parameterized/FinMap/Safe.hs
```diff
@@ -0,0 +1,209 @@
...
+singleton :: a -> FinMap 1 a
+singleton item =
+  FinMap
+    { getFinMap = Map.singleton (Fin.mkFin (NatRepr.knownNat :: NatRepr 0)) item
+    , maxSize = NatRepr.knownNat :: NatRepr 1
+    }
+
+-- | /O(log n)/.
+insert :: Fin n -> a -> FinMap n a -> FinMap n a
+insert k v fm = fm { getFinMap = Map.insert k v (getFinMap fm) }
```
kquick, line 144: Does `maxSize` need to be updated here?

kquick, line 144: Does `maxSize` need to be updated here?

src/Data/Parameterized/FinMap/Safe.hs
```diff
@@ -0,0 +1,209 @@
...
+      case Vec.elemAt k v of
+        Just e -> append k e m
+        Nothing -> incMax m)
+
+------------------------------------------------------------------------
+-- Operations
+
+-- | /O(log n)/.
+delete :: Fin n -> FinMap n a -> FinMap n a
+delete k fm = fm { getFinMap = Map.delete k (getFinMap fm) }
```
kquick, line 183: Does `maxSize` need to be updated here as well?

kquick, line 183: Does `maxSize` need to be updated here as well?

kquick:
No idea why github duplicated all those comments... sorry!

src/Data/Parameterized/FinMap/Safe.hs
```diff
@@ -0,0 +1,209 @@
...
+-- Type
+
+-- | @'FinMap' n a@ is a map with @'Fin' n@ keys and @a@ values.
+data FinMap (n :: Nat) a =
+  FinMap
+    { getFinMap :: Map (Fin n) a
+    , maxSize :: NatRepr n
+    }
+
+instance Eq a => Eq (FinMap n a) where
```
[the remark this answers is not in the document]
langston-barrett: > I suspect this is a common requirement to using this instance, so maybe it would be helpful to add a finMapEq function here that encapsulates that.

To be sure I understand - what would be the type of `finMapEq`, and what would it do? Is it one of the following?
```haskell
finMapEq :: FinMap n a -> FinMap n a -> Bool
finMapEq :: FinMap n a -> FinMap m a -> Bool
```

src/Data/Parameterized/FinMap/Safe.hs
langston-barrett: Yes, or even more directly as `Map.size (getFinMap fm)`, but as you say that's under the assumption that we could construct a `Fin n` from an `Int`, which we can't do safely (there's no way to statically inform GHC that the `Int` is less than `n`). However, you're correct that this inequality holds at runtime, and the unsafe interface takes advantage of that:
```
-- | /O(1)/. Number of elements in the map.
size :: forall n a. FinMap n a -> Fin (n + 1)
size = unsafeFin . IntMap.size . getFinMap
```

src/Data/Parameterized/FinMap/Unsafe.hs
langston-barrett: `FinMap` is a functor, but its value type is `a` rather than `IntMap a`, so this formulation doesn't quite work:
```
src/Data/Parameterized/FinMap/Unsafe.hs:168:12: error:
    • Couldn't match type ‘IntMap a0 -> IntMap a0’ with ‘FinMap n a’
      Expected type: a -> FinMap n a -> FinMap n a
        Actual type: (FinMap n a -> a0)
                     -> FinMap n a -> IntMap a0 -> IntMap a0
    • Probable cause: ‘fmap’ is applied to too few arguments
      In the expression: fmap (IntMap.insert (unsafeFinToInt k))
      In an equation for ‘insert’:
          insert k = fmap (IntMap.insert (unsafeFinToInt k))
    • Relevant bindings include
        k :: Fin n (bound at src/Data/Parameterized/FinMap/Unsafe.hs:168:8)
        insert :: Fin n -> a -> FinMap n a -> FinMap n a
          (bound at src/Data/Parameterized/FinMap/Unsafe.hs:168:1)
    |
168 | insert k = fmap (IntMap.insert (unsafeFinToInt k))
    | 
```

langston-barrett:
> No idea why github duplicated all those comments... sorry!

No worries - thanks for the review! I've resolved the duplicate comments and minor issues (e.g., spelling) that are now fixed and responded to the other comments, so this is ready for the next round.

src/Data/Parameterized/FinMap/Safe.hs
```diff
@@ -0,0 +1,209 @@
...
+
+-- | /O(nlog(n))/. Number of elements in the map.
+size :: forall n a. FinMap n a -> Fin (n + 1)
+size fm =
+  Fin.countFin (maxSize fm) (\k _count -> isJust (lookup (Fin.mkFin k) fm))
+
+------------------------------------------------------------------------
+-- Construction
+
+-- | /O(n log n)/. Increase maximum key/size by 1.
```
kquick, line 111: suggest: "... In other words, this grows the map to allow it to support an additional element.  Note that this is the maximum size; `insert` can be used to add elements to a map that has not reached the maximum size `n`, but to add elements to a "full" map, `incMax` will need to be used fist to grow the map."

src/Data/Parameterized/FinMap/Unsafe.hs
kquick: Ahh, right.

kquick (approved):
Just a couple of additional doc suggestions.

src/Data/Parameterized/FinMap/Safe.hs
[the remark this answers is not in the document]
langston-barrett: Ah, so then we get into some interesting questions - The unsafe interface can't *check* whether `n ~ m`, because it doesn't have a `NatRepr` around. However, both interfaces would be able to check whether the maps had the same (key, value) pairs - modulo that the safe implementation would cast its `Fin n`s to `Natural`s to perform this check, since you also can't compare `Fin n` with `Fin m` directly. I'm open to implementing that - let's see if it comes up in UC-Crux.

src/Data/Parameterized/FinMap.hs
```diff
@@ -0,0 +1,79 @@
...
+    module Data.Parameterized.FinMap.Unsafe
+#else
+    module Data.Parameterized.FinMap.Safe
+#endif
+  ) where
+
+#ifdef UNSAFE_OPS
+import Data.Parameterized.FinMap.Unsafe
+#else
+import Data.Parameterized.FinMap.Safe
```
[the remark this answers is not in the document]
langston-barrett: Here's what the Haddocks look like for this module:

> Documentation
>
> module [Data.Parameterized.FinMap.Unsafe](https://github.com/GaloisInc/parameterized-utils/pull/Data-Parameterized-FinMap-Unsafe.html)

In other words, they consist of a link to whichever implementation has been selected as the default (by defining or not defining `UNSAFE_OPS`).

>  should this include a reference to whichever of the Safe or Unsafe implementations is better documented for individual operations?

I endeavored to give them similar documentation coverage, they also have different complexity on their operations, so the documentation differs in significant ways between them, even though the operations are fundamentally similar.

src/Data/Parameterized/FinMap/Safe.hs
langston-barrett: Done!

PR #134, opened 2022-04-28
FinMap: Traversable instance

langston-barrett:


src/Data/Parameterized/FinMap/Safe.hs
```diff
@@ -73,6 +73,9 @@ instance Foldable (FinMap n) where
   foldMap f = foldMap f . getFinMap
   {-# INLINABLE foldMap #-}
 
+instance Traversable (FinMap n) where
+  traverse f fm = FinMap <$> traverse f (getFinMap fm) <*> pure (maxSize fm)
```
kquick, line 77: FYI, this applicative construction style has quadratic code size (https://www.well-typed.com/blog/2021/12/type-level-sharing-now/).

kquick (approved):

src/Data/Parameterized/FinMap/Safe.hs
langston-barrett: That's very interesting, thanks for sharing the blog post!

Issue #62, opened 2019-09-09
Combine efforts Data.Parameterized.Map and Data.Parameterized.Sum


Ericson2314:
Have you all seen http://hackage.haskell.org/package/dependent-map and http://hackage.haskell.org/package/dependent-sum ? We use and these a bunch at Obsidian and help maintain them. They ostensibly are the same as `Data.Parameterized.Map` `Data.Parameterized.Sum`. It probably makes sense to join efforts in maintaining just one version / see if either of us is missing any extra functionality the other might find useful.

After just a brief scroll, one difference is map equality---specifically the constraint on the values. Your `EqF` does not restrict the type parameter, whereas our http://hackage.haskell.org/package/constraints-extras/docs/Data-Constraint-Extras.html finitely enumerates the possible GADT indices. We are just switching to GHC 8.6, after which I hope it is possible to handle the countably infinite cases that arise in general (with indices and parameters) in that library with quantified constraints, yielding the best of both worlds. This is exactly the nice sort of insight that comes so much more readily from comparing two different implementations of the same idea.

kquick:
Hi @Ericson2314 ,

Thanks for the heads-up about dependent-map and dependent-sum.  It will be interesting to take a look at the similarities between these, but it's going to be something we will need to evaluate over a longer time period since we will need to evaluate the impact of any differences between the libraries.

Out of curiosity, were there aspects of parameterized-utils that were problematic and led you to these alternate libraries?

Ericson2314:
FYI some of this stuff is now been peeled off and upstreamed in the even more minimal https://github.com/phadej/some CC @phadej

PR #142, opened 2022-09-28
RFC: Make `Some` into a newtype around `Some` from package `some`

langston-barrett:
The `Some` type from package `some` is a newtype, and is correspondingly more performant than our previous `data`-based encoding.

This change will require a major version bump and an easy but still mandatory migration for downstream codebases. Downstream code will either need to switch to using the eliminator (`viewSome`), or using pattern synonyms. Example migrations to use pattern synonyms can be seen in this commit.

This patch also uses `-XGeneralizedNewtypeDeriving` to derive the upstream `Semigroup` and `Monoid` instances, which are a good demonstration of the utility of sharing an implementation with a well-thought-out upstream package.

In the long run, I hope this paves the way towards #62. I'm not 100% sure this PR is the right way to go, but I realized that this was possible and thought it was worth a discussion.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,13 @@
     the corresponding `Eq` instances for these data types also require
     `TestEquality` constraints.
 
+  * Make `Some` a newtype around
+    [`Some`](https://hackage.haskell.org/package/some-1.0.4/docs/Data-Some.html#t:Some)
+    from [the `some` package](https://hackage.haskell.org/package/some-1.0.4).
+    This should improve performance, as the latter is a newtype. However, it
+    requires importing the `Some` pattern, rather than a data constructor.
+    Also adds `Semigroup` and `Monoid` instances.
+
 ## 2.1.5.0 -- *2022 Mar 08*
 
   * Add support for GHC 9.2.  Drop support for GHC 8.4 (or earlier).
--- a/parameterized-utils.cabal
+++ b/parameterized-utils.cabal
@@ -62,6 +62,7 @@ library
                , lens           >=4.16 && <5.2
                , mtl
                , profunctors    >=5.6 && < 5.7
+               , some           >=1.0 && < 2.0
                , template-haskell
                , text
                , vector         ==0.12.*
--- a/src/Data/Parameterized/ClassesC.hs
+++ b/src/Data/Parameterized/ClassesC.hs
@@ -15,9 +15,9 @@ Note that there is still some ambiguity around naming conventions, see
 <https://github.com/GaloisInc/parameterized-utils/issues/23 issue 23>.
 -}
 
+{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE Safe #-}
 {-# LANGUAGE TypeOperators #-}
 
 module Data.Parameterized.ClassesC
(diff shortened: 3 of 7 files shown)


RyanGlScott:
Is #62 the primary motivation for this? If so, I wonder if making `parameterized-utils`' `Some` a newtype around the one in `some` is worth it, since the one in `parameterized-utils` would still be distinct. We could, I suppose, eventually switch over to directly using the one in `some`, but that would require yet another breaking change.

The performance benefits from using a newtype encoding of `Some` might be nice, but I'm unclear if this is actually a performance bottleneck in practice.

langston-barrett:
> Is https://github.com/GaloisInc/parameterized-utils/issues/62 the primary motivation for this?

Let me make a more comprehensive pro/con list, as I should have done in the PR description:

**Cons:**

- Major version bump, code churn in dependencies
- Dependency on a new package (possible corresponding increase in build times, lead times when switching to a new GHC, cache sizes, etc.). Counterpoint: We already depend on all of its transitive dependencies, so the impact is likely to be small.

**Pros:**

- Possible performance benefit from newtype representation. Counterpoint: We have no indication that this is a significant performance hit.
- Makes it easier to "import ideas from upstream" such as the `Monoid` instance
- Potential for improved interoperability with `some`, including zero-cost conversions (newtype (un)wrapping) between `Data.Some` and `Data.Parameterized.Some`.
- Potential for an easier migration path towards #62 - downstream packages could individually start migrating to `Data.Some` (with zero-cost conversions to `Data.Parameterized.Some`), and if we can figure out how to migrate typeclasses accordingly we could eventually deprecate `Data.Parameterized.Some`.

RyanGlScott:
OK. If I'm reading the pros correctly, most of the benefits involve the `some` package. If that is true, wouldn't it be more direct to just re-export `some`'s `Some` from `Data.Parameterized.Some`? I'm honestly not sure what benefit it would provide to define our own newtype. If anything, that makes the situation more complicated, since now you have to keep track of which library's `some` you are using.

langston-barrett:
> If that is true, wouldn't it be more direct to just re-export some's Some from Data.Parameterized.Some?

Right, great question. The problem is instances. For example, we have `GEq f => Eq (Data.Some.Some f)`, but `TestEquality f => Eq (Data.Parameterized.Some f)`. It is desireable that we move towards (something like) the former (#129), but doing so would cause *major* downstream breakage and I wanted to start by proposing as minimal of a change as I could. I think we probably should move in the direction of reexporting in the long term, but we would want to have a migration plan in place for these instances.

RyanGlScott:
> The problem is instances. For example, we have `GEq f => Eq (Data.Some.Some f)`, but `TestEquality f => Eq (Data.Parameterized.Some f)`.

Ah, OK.

If we do decide to go down this path, it would be worthwhile to nail down exactly what the desired end state would be. I'm still unclear on what we actually want: Do we want to use a [redefined version of `EqF`](https://github.com/GaloisInc/parameterized-utils/pull/132) in the `Eq` instance for `Some`? Do we want to use `GEq` from `some`? Something else? I'm a bit nervous about committing to this change without having an agreement on the direction.

kquick:
I share the concern about GEq/TestEquality differences (as well as API changes).  I wonder if maybe first we should deprecate `Parameterized.Some` in favor of `some` and let client code selectively switch wherever possible first to minimize the impact.  Ultimately, we might even then just remove `Parameterized.Some` and everywhere use `some`, but if we start from the top rather  than the bottom we can iterate on this and ensure each conversion is valid.

langston-barrett:
Closing due to too many unresolved questions.

Issue #149, opened 2023-02-08
Remove Safe Haskell pragmas


langston-barrett:
I just ran into an issue like #139. `{-# LANGUAGE Safe #-}` annotations seem to cause mysterious and painful-to-debug build failures, and for relatively little benefit. Is there any reason to keep them?

Removing them would probably be considered a breaking change, and require a major version bump.

RyanGlScott:
More concretely, you can reproduce the issue by running:

```
$ cabal repl -b parameterized-utils -w ghc-8.10 --constraint="parameterized-utils==2.1.6.0" --constraint="lens<5"
Warning: The package list for 'hackage.haskell.org' is 23 days old.
Run 'cabal update' to get the latest list of available packages.
Resolving dependencies...
Build profile: -w ghc-8.10.7 -O1
In order, the following will be built (use -v for more details):
 - parameterized-utils-2.1.6.0 (lib) (requires build)
 - fake-package-0 (lib) (first run)
Starting     parameterized-utils-2.1.6.0 (lib)
Building     parameterized-utils-2.1.6.0 (lib)

Failed to build parameterized-utils-2.1.6.0.
Build log (
/home/rscott/.cabal/logs/ghc-8.10.7/parameterized-utils-2.1.6.0-13035a0aac0af162c9e2ac1bd20dbbf2b4de29c08648ef91d6a31e99488888b3.log
):
Configuring library for parameterized-utils-2.1.6.0..
Preprocessing library for parameterized-utils-2.1.6.0..
Building library for parameterized-utils-2.1.6.0..
[ 1 of 38] Compiling Data.Parameterized.Axiom
[ 2 of 38] Compiling Data.Parameterized.Compose
[ 3 of 38] Compiling Data.Parameterized.Classes
[ 4 of 38] Compiling Data.Parameterized.Ctx
[ 5 of 38] Compiling Data.Parameterized.Ctx.Proofs
[ 6 of 38] Compiling Data.Parameterized.DecidableEq
[ 7 of 38] Compiling Data.Parameterized.NatRepr.Internal
[ 8 of 38] Compiling Data.Parameterized.Nonce.Unsafe
[ 9 of 38] Compiling Data.Parameterized.TH.GADT
[10 of 38] Compiling Data.Parameterized.DataKind
[11 of 38] Compiling Data.Parameterized.TraversableFC
[12 of 38] Compiling Data.Parameterized.TraversableF
[...]

PR #150, opened 2023-02-08
Require lens >= 5

### LINKED ISSUE #152: Incorrect lower bound for lens
In addition to the broader suggestion in #149 to remove `Safe` pragmas, there's the more immediate problem that p-u doesn't build with lens < 5, but the Cabal file doesn't say so.

langston-barrett:
Fixes #152.

### DIFF
--- a/parameterized-utils.cabal
+++ b/parameterized-utils.cabal
@@ -59,7 +59,7 @@ library
                , hashable       >=1.2  && <1.5
                , hashtables     >=1.2  && <1.4
                , indexed-traversable
-               , lens           >=4.16 && <5.3
+               , lens           >=5 && <5.3
                , mtl
                , profunctors    >=5.6 && < 5.7
                , template-haskell


RyanGlScott:
In principle, I am supportive of this idea. My only concern about this is that we might make it more difficult to build with other libraries that still require `lens<5`. I don't have a specific example to cite, but I seem to recall that you ran into a situation like this recently?

langston-barrett:
> I don't have a specific example to cite, but I seem to recall that you ran into a situation like this recently?

Yes, unfortunately that's in a closed-source project so I can't share specifics.

> My only concern about this is that we might make it more difficult to build with other libraries that still require lens<5.

Doesn't your example on #149 demonstrate that it's already impossible to build p-u with libraries that depend on lens<5? Or am I misunderstanding that?

FWIW, lens 5 came out a full two years ago.

RyanGlScott:
> Doesn't your example on https://github.com/GaloisInc/parameterized-utils/issues/149 demonstrate that it's already impossible to build p-u with libraries that depend on lens<5?

Currently, that is the case. The question is whether we would want to support doing so into the future. Given how old `lens-5` is, however, perhaps the solution is to just force the rest of the ecosystem to catch up.

langston-barrett:
After merging #151 and so fixing #152, there is no immediate need to change this lower bound.

PR #151, opened 2023-02-08
Mark `ClassesC` as Trustworthy

### LINKED ISSUE #152: Incorrect lower bound for lens

langston-barrett:
For compatibility with lens < 5. Fixes #152.

### DIFF
--- a/parameterized-utils.cabal
+++ b/parameterized-utils.cabal
@@ -43,6 +43,7 @@ common bldflags
                -Werror=incomplete-patterns
                -Werror=missing-methods
                -Werror=overlapping-patterns
+               -Wno-trustworthy-safe
                -fhide-source-paths
   default-language: Haskell2010
 
--- a/src/Data/Parameterized/ClassesC.hs
+++ b/src/Data/Parameterized/ClassesC.hs
@@ -17,7 +17,7 @@ Note that there is still some ambiguity around naming conventions, see
 
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE Safe #-}
+{-# LANGUAGE Trustworthy #-}
 {-# LANGUAGE TypeOperators #-}
 
 module Data.Parameterized.ClassesC


RyanGlScott:
GHC [now warns](https://github.com/GaloisInc/parameterized-utils/actions/runs/4126215984/jobs/7127838429#step:11:617):

```
[20 of 38] Compiling Data.Parameterized.ClassesC
src/Data/Parameterized/ClassesC.hs:20:14: warning: [-Wtrustworthy-safe]
Warning:     ‘Data.Parameterized.ClassesC’ is marked as Trustworthy but has been inferred as safe!
   |
20 | {-# LANGUAGE Trustworthy #-}
   |              ^^^^^^^^^^^
```

I think we should just globally set `-Wno-trustworthy-safe` in the `.cabal` file to avoid this class of warnings.

langston-barrett:
Good catch, fixed!

RyanGlScott:
I think this is a sensible way to handle things. Does this work for you, @kquick?

kquick:
Couldn't `subterms someone something` internally call `unsafePerformIO` and wouldn't this violate the intention of declaring this module `Trustworthy`?

RyanGlScott:
My understanding of the contract that `Trustworthy` offers (as documented [here](https://downloads.haskell.org/ghc/9.4.4/docs/users_guide/exts/safe_haskell.html#trustworthy-requirements)) is that `Trustworthy` code _is_ allowed to do unsafe things internally, provided that the user-facing API still respects type safety and referential transparency. That is certainly the case with `Data.Parameterized.ClassesC`, and I'd wager the same applies to nearly all the code in `parameterized-utils` (with the possible exception of [`Data.Parameterized.Axiom`](https://github.com/GaloisInc/parameterized-utils/blob/4ed9f5c64a6858858cf8545b92ccc73eee2d21f3/src/Data/Parameterized/Axiom.hs)).

kquick:
OK.  I have reservations about who bears the responsibility on `Trustworthy`, but I think that's outside the scope of what is reasonably involved here, so I'm good with this change.

RyanGlScott (approved):

RyanGlScott:
Are you alright with merging this, @langston-barrett?

PR #153, opened 2023-05-24
Support building with GHC 9.6

RyanGlScott:
The primary change, which is required to make `parameterized-utils` compile with GHC 9.6, is:

# Avoid clashing with `GHC.TypeNats`' `withKnownNat`

`base-4.18.*` (GHC 9.6.*) exports a variety of singleton-related functionality from `GHC.TypeNats` (see [this
section](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.6?version_id=e9996b225b1f2a5005178283f833fe6ce341beb3#knownnat-knownsymbol-knownchar-changes) of the GHC 9.6 Migration Guide). This includes a `withKnownNat` function that clashes with the one defined in `Data.Parameterized.NatRepr`. This patch uses explicit imports from `GHC.TypeNats` in that module to avoid this issue.

-----

I have also included a number of commits to fix warnings introduced in `-Wall` in GHC 9.6. These include the fact that `liftA2` is now re-exported from the `Prelude`, which requires making some `Control.Applicative` imports qualified to avoid redundant import warnings, as well as `TypeInType` being deprecated as of GHC 9.6, which requires guarding its use with CPP.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,8 +51,8 @@ jobs:
             8.6.5) GHC_NIXPKGS=github:nixos/nixpkgs/20.09 ;;
             9.0.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-22.05 ;;
             9.2.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-22.05 ;;
-            9.4.3) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-unstable ;;
-            9.6.1) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-unstable ;;
+            9.4.5) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-23.05 ;;
+            9.6.1) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-23.05 ;;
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -52,6 +52,7 @@ jobs:
             9.0.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-22.05 ;;
             9.2.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-22.05 ;;
             9.4.3) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-unstable ;;
```
kquick, line 54: We can now reference tagged `nixpkgs/nixos-23.05` for GHC 9.4.5  (while you're here).  We could also get 9.6.1 from that same tag, although continuing to use nixos-unstable will get bugfix revisions in the 9.6 series... your call.

langston-barrett (approved):

kquick (approved):

.github/workflows/test.yml
RyanGlScott: I've converted both of these over to `nixos-23.05`. If we want to upgrade the 9.6 minor version in the future, we can always switch to using `nixos-unstable`.

PR #154, opened 2023-06-07
Typed structural equality

kquick:
The structuralEquality TH implementation does not witness the equality of the types involved, which can make it difficult to use for parameterized GADT types.  It also cannot correctly generate the comparison slices for parameterized GADTs because there may not be wildcard cases admissable due to type equality, even if there are other constructors with *other* resulting types.

The impact should be minimal, although it may require the introduction of the `ScopedTypeParameters` pragma in sources importing and using this code.

### DIFF
--- a/src/Data/Parameterized/TH/GADT.hs
+++ b/src/Data/Parameterized/TH/GADT.hs
@@ -257,6 +257,15 @@ mkSimpleEqF dTypes bnd pats con xv yQ multipleCases argsSameType = do
   (yp,yv) <- conPat con "y"
   let rv = matchEqArguments dTypes pats nm bnd (constructorFields con) xv yv
   let otherMatchingCons =
+        -- Determine the other constructors that should be matched relative to
+        -- `con`.  If this is supplying code for `testEquality`, the input
+        -- signature is `f a -> f b -> ...` and will admit different types, so
+        -- all constructors should be checked, but if this is supplying code for
+        -- `Eq` or similar where the input signature is `a -> a -> ...`
+        -- (i.e. `argsSameType` is `True`), then only constructors that have the
+        -- same resulting type should be checked, otherwise GHC will emit
+        -- warnings/errors about "pattern not reached" for the case statement
+        -- being generated here.
         let sameContext = (==) `on` constructorContext
         in if argsSameType
            then filter (sameContext con) multipleCases


RyanGlScott (commented):
Reading the PR description, it wasn't obvious to me why this change was necessary in the first place. Here is my attempt to describe the issue, after looking at the `-ddump-splices` output a bit.

When you use `structuralTypeEquality`, like in this example:

```hs
data T a where
  MkT1 :: T Int
  MkT2 :: T Bool

instance TestEquality T where
  testEquality :: T a -> T b -> Maybe (a :~: b)
  testEquality = $(structuralTypeEquality [t|T|] []) 
```

Then you will generate code involving GADT pattern-matching. This can interact poorly with type inference in general, but because the type of `testEquality` mentions `Maybe (a :~: b)` in its return type, this provides enough information to allow the GADT pattern matches to type-check.

`structuralEquality` is a minor variation of `structuralTypeEquality`. If you write this:

```hs
instance TestEquality T where
  testEquality :: T a -> T a -> Bool
  testEquality = $(structuralEquality [t|T|] []) 
```

It is equivalent to writing this:

```hs
instance TestEquality T where
  testEquality :: T a -> T a -> Bool
  testEquality = isJust $(structuralTypeEquality [t|T|] [])
```

This is very similar to the previous example, but with a key difference: the return type is now `Bool`, which no longer mentions any of the type variables. As a result, GHC's type inference has no way to know what the overall type of the generated GADT pattern matches should be, and therefore the generated code will fail to type-check. (Concretely, GHC will complain that that the `a` in `T a` is untouchable.)

One way to fix this issue is to add an explicit type ascription to the generated `case` expression that makes the return type of the `case` clear. This requires binding type variables, as they may not yet be in scope:

```hs
instance TestEquality T where
  testEquality :: T a -> T a -> Bool
  testEquality = isJust (\arg1 arg2 -> (\(x :: T a) (y :: T b) -> case (x, y) of { ... } :: Maybe (a :~: b)) arg1 arg2)
```

This is the approach taken in this PR.

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -133,10 +135,72 @@ typeVars :: TypeSubstitution a => a -> Set Name
...
+  -- tpq is some type of GADT: data X p1 p2 ... where ...
+  --
+  -- The general approach is to generate a structural type equality such that the
+  -- result is a Maybe (e :+: f) is Just Refl and then verify it is a Just value
+  -- to assert equality by generating (via template haskell):
+  --
+  --   \ x y -> isJust $(structuralTypeEquality ... x y)
+  --
+  -- However, that result presumes a `TestEquality f where testEquality :: f a ->
+  -- f b -> Maybe (a :~: b)`.  If the GADT has a single type parameter, those
```
langston-barrett, line 152: It looks to me like `structuralEquality` calls `structuralTypeEquality`, which generates (the syntax of) a function of type `f a -> f b -> Maybe (a :~: b)`, meaning that using `structuralEquality` would *not* require that `tpq` have an instance of `TestEquality`.

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -133,10 +135,72 @@ typeVars :: TypeSubstitution a => a -> Set Name
...
+  -- pragma, but GHC will happily suggest that if it's missing.
+  --
+  -- This is also useful for the equality test on the single parameter case:
+  --
+  --   data D p1 where ...
+  --
+  --   instance Eq (D a) where
+  --     (==) = $(structuralEquality [t|D|] []
+  --
+  -- Again, this will fail without the template haskell assertion of the target
```
langston-barrett, line 178: To me, "assertion" implies "checked (and may fail) at run-time". I could see "ascription" or "annotation" here.
```suggestion
  -- Again, this will fail without the template haskell ascription of the target
```

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -133,10 +135,72 @@ typeVars :: TypeSubstitution a => a -> Set Name
...
+  arg1Params <- fmap varT <$> newNames "xTy" (length gadtParams)
+  arg2Params <- fmap varT <$> newNames "yTy" (length gadtParams)
+  let arg1Ty = foldl appT (conT $ datatypeName d) arg1Params
+  let arg2Ty = foldl appT (conT $ datatypeName d) arg2Params
+#if MIN_VERSION_base(4,14,0)
+  let mkSuperTy tyList = foldl appT (promotedTupleT (length tyList)) tyList
+#else
+  let mkSuperTy tyList =
+         if length tyList < 2
+         then if length tyList == 0
```
langston-barrett, line 191: I'm not sure how much computation is really done, i.e., how lazy length is, but this form makes it abundantly clear that it is not at all necessary to compute the length:
```suggestion
         then if null tyList
```

src/Data/Parameterized/TH/GADT.hs
```diff
@@ -133,10 +135,72 @@ typeVars :: TypeSubstitution a => a -> Set Name
...
+              then error "Expected at least one type in structuralEquality"
```
langston-barrett, line 192: This error would be for cases like `$(structuralTypeEquality [t|Bool])`, right? Perhaps this would be more clear?
```suggestion
              then error "Expected at least one type parameter in structuralEquality"
```

test/Test/TH.hs
```diff
@@ -39,13 +41,41 @@ instance TestEquality T2Repr where
...
 eqTest :: (TestEquality f, Show (f a), Show (f b)) => f a -> f b -> IO ()
 eqTest a b =
   when (not (isJust (testEquality a b))) $ assertFailure $ show a ++ " /= " ++ show b
 
 neqTest :: (TestEquality f, Show (f a), Show (f b)) => f a -> f b -> IO ()
 neqTest a b =
-  when (isJust (testEquality a b)) $ assertFailure $ show a ++ " == " ++ show b
+  when (isJust (testEquality a b))
+  $ assertFailure
+  $ show a <> " == " <> show b <> " but should not be!"
```
langston-barrett, line 72: nit: Looks like these are still strings, so I'd prefer keeping `++` rather than `<>`, as it's less polymorphic, and so lets the reader know the more specific types involved.

test/Test/TH.hs
```diff
@@ -62,6 +92,28 @@ thTests = testGroup "TH" <$> return
...
+      T3_Int 1 `neqTest` T3_Int 3
+      T3_Int 1 `neqTest` T3_Bool True
+      T3_Bool False `neqTest` T3_Bool True
+      T3_Bool True `eqTest` T3_Bool True
+
+      assertEqual "T4_Int values" (T4_Int @String 5) (T4_Int @String 5)
+      assertNotEqual "T4_Int values" (T4_Int @String 5) (T4_Int @String 54)
+
+      T4_Int @String 1 `eqTest` T4_Int @String 1
+      T4_Int @String 1 `neqTest` T4_Int @String 2
```
langston-barrett, line 114: Could maybe get more coverage with some property tests:
```
forall (i j :: Int). (T3_Int i == T3_Int j) == (i == j)
[same for Bool]
forall (i j :: Int). (T3_Int i == T3_Int j) == isJust (testEquality (T3_Int i) (T3_Int j))
...
```

PR #173, opened 2025-02-13
ci: Enable `-Werror=compat -Werror=default`

langston-barrett:
To prevent warnings sneaking in without us noticing! Also, switch to a non-deprecated Haskell setup action to get a non-ancient version of Cabal.

I don't feel strongly about this, just thought it would be nice and figured we could do it if we all agreed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -78,7 +78,8 @@ jobs:
         run: cabal check
       - name: Configure
         shell: bash
-        run: cabal configure --enable-tests
+        # See doc/dev.md for development practices around warnings.
+        run: cabal configure --enable-tests --ghc-options='-Werror=compat -Werror=default'
       - name: Build
         shell: bash
         run: cabal build
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -2,14 +2,21 @@
 
 ## Warnings
 
-`parameterized-utils` is a research-grade codebase, so correctness in all cases is not *always* the highest priority.
-In particular, as developers, we must weigh the cost of development practices that prioritize correctness, uniformity, or other concerns against our desire to encourage relatively rapid prototyping of interesting functionality.
+`parameterized-utils` is a research-grade codebase.
+As developers, we should therefore adopt development practices that enable relatively rapid and frictionless prototyping of interesting functionality.
+However, correctness remains a major concern, and we should be willing to invest some effort in practices that reduce bugs.
 When forming judgements about appropriate practices, it is worth remembering that `parameterized-utils` is also upstream of a substantial number of projects, such as What4, Crucible, and Macaw.
 
 The developers currently judge that the warnings included in GHC's `-Wdefault` are conservative enough to be worth fixing in most cases.
-Accordingly, the current default development practice is to fix instances of `-Wdefault` in new code.
+The documentation for `-Wdefault` [states](https://downloads.haskell.org/ghc/latest/docs/users_guide/using-warnings.html#ghc-flag-Wdefault) that they are:
+
+> a standard set of warnings which are generally likely to indicate bugs in your program[^wdefault]
+
+Given that correctness is a high priority, the current default development practice is to fix instances of `-Wdefault` in new code.
 However, we also trust one anothers' judgement to override this practice in any particular case.
 When doing so, developers are encouraged to disable individual warnings on a per-module basis, ideally accompanied by a comment including some justification for why the warnings were not (and/or should not be) fixed.
 To prevent warnings from slipping in unnoticed and thus unexamined, we enable `-Werror=default` in CI.
 
 We also enable `-Werror=compat` in order to gradually prepare for breaking changes in GHC.
+
+[^wdefault]: Unfortunately, this is not strictly true. `-Wdefault` includes (for example) `-Wtabs`, which is about code style, and `-Winaccessible-code`, which is about dead code. However, the majority of the warnings in `-Wdefault` do indeed affect correctness, and hence should generally be fixed.
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -19,4 +19,9 @@ To prevent warnings from slipping in unnoticed and thus unexamined, we enable `-
 
 We also enable `-Werror=compat` in order to gradually prepare for breaking changes in GHC.
 
+To reproduce the behavior of the CI build locally, developers can run
+```sh
+cabal configure --enable-tests --ghc-options='-Werror=compat -Werror=default'
+```
+
 [^wdefault]: Unfortunately, this is not strictly true. `-Wdefault` includes (for example) `-Wtabs`, which is about code style, and `-Winaccessible-code`, which is about dead code. However, the majority of the warnings in `-Wdefault` do indeed affect correctness, and hence should generally be fixed.
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -1,0 +1,15 @@
+# Developer documentation
+
+## Warnings
+
+`parameterized-utils` is a research-grade codebase, so correctness in all cases is not *always* the highest priority.
+In particular, as developers, we must weigh the cost of development practices that prioritize correctness, uniformity, or other concerns against our desire to encourage relatively rapid prototyping of interesting functionality.
+When forming judgements about appropriate practices, it is worth remembering that `parameterized-utils` is also upstream of a substantial number of projects, such as What4, Crucible, and Macaw.
+
+The developers currently judge that the warnings included in GHC's `-Wdefault` are conservative enough to be worth fixing in most cases.
+Accordingly, the current default development practice is to fix instances of `-Wdefault` in new code.
+However, we also trust one anothers' judgement to override this practice in any particular case.
+When doing so, developers are encouraged to disable individual warnings on a per-module basis, ideally accompanied by a comment including some justification for why the warnings were not (and/or should not be) fixed.
+To prevent warnings from slipping in unnoticed and thus unexamined, we enable `-Werror=default` in CI.
+
+We also enable `-Werror=compat` in order to gradually prepare for breaking changes in GHC.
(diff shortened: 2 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         run: cabal check
       - name: Configure
         shell: bash
-        run: cabal configure --enable-tests
+        run: cabal configure --enable-tests --ghc-options='-Wall -Wcompat -Werror'
```
RyanGlScott, line 81: I think you only need to enable `-Werror` here, since both `-Wall` and `-Wcompat` are already enabled in the `.cabal` file.

src/Data/Parameterized/ClassesC.hs
```diff
@@ -20,6 +20,9 @@ Note that there is still some ambiguity around naming conventions, see
 {-# LANGUAGE Trustworthy #-}
 {-# LANGUAGE TypeOperators #-}
 
+-- See https://github.com/GaloisInc/parameterized-utils/issues/149
+{-# OPTIONS_GHC -Wno-trustworthy-safe #-}
```
RyanGlScott, line 24: Personally, `-Wtrustworthy-safe` is among the warnings that I would elect to disable globally, since I rarely find the amount of effort it causes to fix these warnings to be worth the hassle. (I'm fine with just disabling it per-module as is done here, however.)

RyanGlScott (commented):
I don't have a problem with `-Werror` myself, but I reserve the right to complain if I find a particular class of warning to be annoying. In fact, this PR already goes as far as disabling a warning in `Data.Parameterized.ClassesC`, so I think you're on the same page as I am :)

src/Data/Parameterized/ClassesC.hs
langston-barrett: Not only is it not *worth* fixing, it's actually *problematic* to fix, as seen in #149...!

langston-barrett:
One point to consider would be whether `-Wall -Wcompat` is the right set to error on. As Ryan noted above, `-Wall` actually has some questionable inhabitants. Perhaps we should just `-Werror=compat -Werror=default`? The warnings in `default` are somewhat more conservative, while still containing most of `all`.

src/Data/Parameterized/ClassesC.hs
langston-barrett: I would vote for not accepting any more modules with `Trustworthy` nor `Safe` markers in the first place.

RyanGlScott:
> Perhaps we should just `-Werror=compat -Werror=default`?

I'd be fine with that.

sauclovian-g:
I don't have a strong opinion, other than the warning about unused values, which IME is critical for avoiding accidental loops. (I could probably have an opinion if I sat down with the manual and the list of warnings; LMK if it would be useful for me to do that. As it is, the last time I did that was in I think GHC 6.x.)

sauclovian-g:
The reason I like -Werror is that if there are expected warnings, it's very easy for other unexpected warnings to slip in unnoticed. (There are other ways to attend to this concern, but they're usually more expensive...)

langston-barrett:
> Also, lets make a separate PR for the action version bump, since it's unrelated to the rest of this PR and associated discussions.

Good call, https://github.com/GaloisInc/parameterized-utils/pull/177

src/Data/Parameterized/ClassesC.hs
langston-barrett: Switched to `-Wdefault`, which does not include this warning. Left this line in, because `-Wall` is already enabled in the Cabal file.

langston-barrett:
> To put it another way, if I didn't care about these warnings enough to fix them during my development, are they a required must-fix before sharing this code?

Completely agreed that if code with warnings is *intentionally* included, we should trust one anothers' judgement over the compiler's. Also agreed with @sauclovian-g that it is hard to reliably keep track of which warnings were intentionally ignored vs. unintentionally added without some automated assistance (especially reliably, especially with multiple developers, especially over long timespans).

> I think we should have a position on the development requirements for this codebase, and those should be documented (probably in a Development.md to avoid bloating the README.md). I'm definitely open to discussion regarding my proposal of what our position should be above, but would request that this PR be supplemented with the documentation of this position for the benefit of ourselves, future maintainers, and users. I think this also becomes relevant for those future maintainers when a future compiler version introduces a new Werror failure.

Great point, I've added some documentation. It mentions some of the concerns mentioned above (i.e., correctness vs. rapid/frictionless prototyping), and one not yet mentioned above: there are downstream codebases that have their own concerns.

Let me know what you think! Thanks for the thoughtful feedback, all!

.github/workflows/test.yml
```diff
@@ -78,7 +78,8 @@ jobs:
         run: cabal check
       - name: Configure
         shell: bash
-        run: cabal configure --enable-tests
+        # See doc/dev.md for development practices around warnings.
+        run: cabal configure --enable-tests --ghc-options='-Werror=compat -Werror=default'
```
sauclovian-g, line 82: I would suggest moving this to the cabal config so it happens on all builds, not just the CI builds. IMO at least it's really irritating to think you're done, push a branch and open a pull request, and then discover it doesn't even compile. (Especially if by Murphy's Law it then turns out you're only half done with the changes you need.)

doc/dev.md
```diff
@@ -0,0 +1,15 @@
+# Developer documentation
+
+## Warnings
+
+`parameterized-utils` is a research-grade codebase, so correctness in all cases is not *always* the highest priority.
```
kquick, line 5: Thank you for this writeup!

I think it would be good to define "correctness" here, since I think we would all agree that correctness in terms of proper output is a priority, but what we are talking about here is compilation concerns and so is probably better referred to as "code hygene", or with a good reference to what type of "correctness" is being discussed.

.github/workflows/test.yml
langston-barrett: > IMO at least it's really irritating to think you're done, push a branch and open a pull request, and then discover it doesn't even compile.

I completely agree. However, if this goes in the `.cabal` file, then it can cause builds to fail with newer GHC versions if they add new warnings to `-Wcompat` or `-Wdefault`, which can be annoying for downstream projects that want to update GHC versions. We already have [`-Wall` and `-Wcompat` in the `.cabal` file](https://github.com/GaloisInc/parameterized-utils/blob/8a349603b686a2a51d2b6eeb90458890fc86e310/parameterized-utils.cabal#L39-L40), so at the very least, developers see a superset of these warnings locally (even if they don't cause the local build to fail).

In many projects, I simply list all of the warnings I want to fail the build explicitly, enabled for the GHC versions for which they apply. See further discussion and such a list [here](https://github.com/GaloisInc/crucible/blob/90447535c05834fcb2bd7bb0aa64847b788f69bf/crucible-debug/crucible-debug.cabal#L18). I update this list when new GHC versions come out. `parameterized-utils` has a much smaller list [here](https://github.com/GaloisInc/parameterized-utils/blob/8a349603b686a2a51d2b6eeb90458890fc86e310/parameterized-utils.cabal#L43-L45).

doc/dev.md
langston-barrett: Ah, this is a helpful point of clarification - I actually *did* mean correctness in terms of proper output, as `-Wdefault` includes warnings such as `-Wmissing-methods` and `-Woverlapping-patterns`, and I would say that the presence of such a warning often indicates a "correctness in terms of proper output"-type bug.

However, I agree that some members of `-Wdefault` do *not* concern this type of correctness, such as `-Wtabs`.

I will make this distinction more clear.

langston-barrett: Pushed another commit that highlights correctness as a major concern, and mentions that the purpose of `-Wdefault` is to reduce bugs, though also includes a footnote on ways in which that is not quite true in practice.

kquick: Looks good.  And yes, I frequently turn on the warnings you mentioned in other projects because of the implications, whilst not usually going as far as  `-Werror -Wdefault`.

kquick (approved):

langston-barrett:
> One point to consider would be whether -Wall -Wcompat is the right set to error on. As Ryan noted above, -Wall actually has some questionable inhabitants. Perhaps we should just -Werror=compat -Werror=default? The warnings in default are somewhat more conservative, while still containing most of all.

> other than the warning about unused values, which IME is critical for avoiding accidental loops. 

I just noticed that `-Wunused-binds` is actually not in `-Wdefault`, but only in `-Wextra`... As noted above, I tend to land on the more aggressive side of warnings-enforcement. I usually enable `-Werror` for all of `-Wall` (plus a few other things, minus `-Wtrustworthy-safe`) so I would *a fortiori* also be fine with `-Werror=extra`. @kquick noted that he generally doesn't go as far as `-Werror=default`, so presumably would be *less* enthused about `-Wextra`. @sauclovian-g indicated in the quote above that he doesn't really care about the particular set except that it should specifically include `-Wunused-binds`... I'm worried that we might be on the verge of bike-shedding here :sweat_smile: Rather than turn on `-Werror=extra` or even `-Werror=unused-binds`, I propose we appeal to authority (a [classic fallacy](https://en.wikipedia.org/wiki/Argument_from_authority), but perhaps helpful to avoid bikeshedding here), keep `-Werror=default`, and trust the GHC developers to maintain `-Wdefault` as a reasonable list.

.github/workflows/test.yml
sauclovian-g: My feeling about that (since with C code and gcc it happens unpredictably in ways you can't prevent, because a lot of gcc warnings depend on details of the optimizer behavior) is that staying on top of new compiler versions is just a necessary maintenance cost.

But also, this is why a lot of projects have "maintainer mode" builds where that's what engages -Werror. Is that possible with cabal?

sauclovian-g: > Would you like me to add that to doc/dev.md?

I think that covers most of the concerns, so yeah that would be nice.

langston-barrett: Done! Thanks for raising this concern, I agree that it is unfortunately a bit of an annoyance.

RyanGlScott (approved):

sauclovian-g (approved):

kquick (approved):

PR #178, opened 2025-03-05
doc: Describe GHC version support policy

langston-barrett:
Also add instructions on how to add or remove GHC versions. I find this helpful because the steps can be different for different repos (e.g., repos may or may not have a `tested-with` field, Cabal freeze files, Dockerfiles, etc.).

I have no particular commitment to the three-version policy, it's just the de facto standard amongst our projects.

### DIFF
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -1,5 +1,27 @@
 # Developer documentation
 
+## GHC versions
+
+We support at least three versions of GHC at a time.
+Older versions may be dropped from CI.
+We try to support new versions as soon as they are supported by all of libraries that we depend on.
+
+### Adding a new version
+
+The following checklist enumerates the steps needed to support a new version of GHC.
+When performing such an upgrade, it can be helpful to copy/paste this list into the MR description and check off what has been done, so as to not forget anything.
+
+- [ ] Allow the new version of `base` in the Cabal `build-depends`
+- [ ] Fix any new warnings from `-Wdefault`
+- [ ] Add the new GHC version to the matrix in the Github Actions configuration
+- [ ] Add the new GHC version to the Cabal `tested-with` field
+- [ ] Optionally follow the below steps to remove any old GHC versions
+
+### Removing an old version
+
+- [ ] Remove the old version from the matrix in the Github Actions configuration
+- [ ] Remove the old version from the Cabal `tested-with` field
+
 ## Warnings
 
 `parameterized-utils` is a research-grade codebase.
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -13,13 +13,13 @@ When performing such an upgrade, it can be helpful to copy/paste this list into 
 
 - [ ] Allow the new version of `base` in the Cabal `build-depends`
 - [ ] Fix any new warnings from `-Wdefault`
-- [ ] Add the new GHC version to the matrix in the Github Actions configuration
+- [ ] Add the new GHC version to the matrix in the GitHub Actions configuration
 - [ ] Add the new GHC version to the Cabal `tested-with` field
 - [ ] Optionally follow the below steps to remove any old GHC versions
 
 ### Removing an old version
 
-- [ ] Remove the old version from the matrix in the Github Actions configuration
+- [ ] Remove the old version from the matrix in the GitHub Actions configuration
 - [ ] Remove the old version from the Cabal `tested-with` field
 
 ## Warnings
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -3,7 +3,7 @@
 ## GHC versions
 
 We support at least three versions of GHC at a time.
-We are not aggressive about dropping older versions, but will generally do so for versions outside of the support window if maintaining that support would require efforts such as C pre-processor `ifdef` sections or Cabal `if` sections.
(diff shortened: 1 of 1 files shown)


doc/dev.md
```diff
@@ -1,5 +1,27 @@
...
+
+- [ ] Allow the new version of `base` in the Cabal `build-depends`
+- [ ] Fix any new warnings from `-Wdefault`
+- [ ] Add the new GHC version to the matrix in the Github Actions configuration
+- [ ] Add the new GHC version to the Cabal `tested-with` field
+- [ ] Optionally follow the below steps to remove any old GHC versions
+
+### Removing an old version
+
+- [ ] Remove the old version from the matrix in the Github Actions configuration
```
RyanGlScott, line 22: I swear I am not one of GitHub's lawyers:

```suggestion
- [ ] Add the new GHC version to the matrix in the GitHub Actions configuration
- [ ] Add the new GHC version to the Cabal `tested-with` field
- [ ] Optionally follow the below steps to remove any old GHC versions

### Removing an old version

- [ ] Remove the old version from the matrix in the GitHub Actions configuration
```

RyanGlScott (approved):

doc/dev.md
```diff
@@ -1,5 +1,27 @@
 # Developer documentation
 
+## GHC versions
+
+We support at least three versions of GHC at a time.
+Older versions may be dropped from CI.
+We try to support new versions as soon as they are supported by all of libraries that we depend on.
```
sauclovian-g, line 7: typo: s/all of libraries/all of the libraries/ (or "all the libraries")

sauclovian-g (approved):

doc/dev.md
```diff
@@ -1,5 +1,27 @@
 # Developer documentation
 
+## GHC versions
+
+We support at least three versions of GHC at a time.
+Older versions may be dropped from CI.
```
kquick, line 6: Consider adding: "We are not aggressive about dropping older versions, but will generally do so for versions outside of the support window if maintaining that support would require considerable backward-maintenance changes such as significant numbers of CPP ifdef sections or cabal `if` sections."

doc/dev.md
```diff
@@ -1,5 +1,27 @@
...
+- [ ] Allow the new version of `base` in the Cabal `build-depends`
+- [ ] Fix any new warnings from `-Wdefault`
+- [ ] Add the new GHC version to the matrix in the GitHub Actions configuration
+- [ ] Add the new GHC version to the Cabal `tested-with` field
+- [ ] Optionally follow the below steps to remove any old GHC versions
+
+### Removing an old version
+
+- [ ] Remove the old version from the matrix in the GitHub Actions configuration
+- [ ] Remove the old version from the Cabal `tested-with` field
```
kquick, line 23: Consider adding:
```
[ ] Check the code for any CPP ifdefs that refer to the dropped version and make the associated changes.
[ ] Check the cabal file for any conditional sections based on the dropped version and eliminate the conditional as appropriate.
```

kquick (commented):
Thanks for putting this together!  I had a couple of suggestions for additions; happy to discuss if you have any questions or concerns.

doc/dev.md
```diff
@@ -1,5 +1,27 @@
...
+We try to support new versions as soon as they are supported by all of libraries that we depend on.
+
+### Adding a new version
+
+The following checklist enumerates the steps needed to support a new version of GHC.
+When performing such an upgrade, it can be helpful to copy/paste this list into the MR description and check off what has been done, so as to not forget anything.
+
+- [ ] Allow the new version of `base` in the Cabal `build-depends`
+- [ ] Fix any new warnings from `-Wdefault`
+- [ ] Add the new GHC version to the matrix in the GitHub Actions configuration
```
kquick, line 16: I plan to extract the configuration text via nix to a standalone action, but until then, that should be updated for the new version as well (e.g.  https://github.com/GaloisInc/parameterized-utils/blob/master/.github/workflows/test.yml#L57).

doc/dev.md
langston-barrett: Nice, thanks! This is why we need checklists - I would clearly forget this part otherwise :)

doc/dev.md
kquick: I'd rather not remove support for older versions unless it's particularly onerous to maintain.  Some of our customers are stuck using very old OS versions with corresponding packages.

kquick: Sorry, I release the tail end of the last comment was a bit opaque.  Let me try again: The OS version likely dictates the GHC version available, but it would be nice if the GHC version didn't limit the versions of Haskell-based applications and their dependencies that  we could successfully build.

kquick (approved):

doc/dev.md
langston-barrett: > I'd rather not remove support for older versions unless it's particularly onerous to maintain. Some of our customers are stuck using very old OS versions with corresponding packages.

Ah, this is valuable context! In that case, I wonder if we should consider widening the 3-version guaranteed support window, for this package and other Galois packages?

langston-barrett:
There's a bit of an ongoing conversation, but it looks like we have enough consensus to treat this as a first draft.

doc/dev.md
RyanGlScott: [`copilot-theorem`](https://hackage.haskell.org/package/copilot-theorem) is one example of a library that aims to have broader support than only three releases. The most it depends on is `what4`. For other libraries, (e.g., Crucible), I think it would be fine to keep the 3-release policy pending further needs.

PR #181, opened 2025-03-13
ci: Small improvements

langston-barrett:
See individual commits.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
         allow-failure: [false]
       fail-fast: false
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -9,12 +9,13 @@ jobs:
     continue-on-error: ${{ matrix.allow-failure }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-22.04]
```
RyanGlScott, line 12: Perhaps we should keep this at `ubuntu-24.04`? That's the current value of `ubuntu-latest`, and there's nothing else that would constrain us to `ubuntu-22.04`.

RyanGlScott (approved):

Issue #180, opened 2025-03-11
Fix Haddock warnings


langston-barrett:
Running `cabal haddock` generates dozens of warnings. Perhaps we should fix these and enable some more strict form of checking for them in CI?

A sampling:
```
arning: 'fmapF' is out of scope.
    If you qualify the identifier, haddock can try to link it anyway.
Warning: 'FunctorF' is out of scope.
    If you qualify the identifier, haddock can try to link it anyway.
  95% ( 20 / 21) in 'Data.Parameterized.TraversableFC'
  Missing documentation for:
    TraversableFC (src/Data/Parameterized/TraversableFC.hs:185)
  93% ( 15 / 16) in 'Data.Parameterized.TraversableF'
  Missing documentation for:
    TraversableF (src/Data/Parameterized/TraversableF.hs:156)
Warning: 'Some' is ambiguous. It is defined
    * at src/Data/Parameterized/Some.hs:30:29
    * at src/Data/Parameterized/Some.hs:30:1
    You may be able to disambiguate the identifier by qualifying it or
    by specifying the type/value namespace explicitly.
    Defaulting to the one defined at src/Data/Parameterized/Some.hs:30:1
  85% (  6 /  7) in 'Data.Parameterized.Some'
  Missing documentation for:
    Some (src/Data/Parameterized/Some.hs:30)
Warning: 'SymbolRepr' is ambiguous. It is defined
    * at src/Data/Parameterized/SymbolRepr.hs:53:5
    * at src/Data/Parameterized/SymbolRepr.hs:52:1
    You may be able to disambiguate the identifier by qualifying it or
    by specifying the type/value namespace explicitly.
    Defaulting to the one defined at src/Data/Parameterized/SymbolRepr.hs:52:1
```
https://github.com/GaloisInc/parameterized-utils/actions/runs/13793015551/job/38577571862#step:15:13

PR #184, opened 2025-03-14
Clean up Haddocks

langston-barrett:
Progress towards #180.

`WithRepr`:

- Put some identifiers that should not be linked in `@` rather than `'`
- Don't explicitly note that a method with a default implementation doesn't need to be implemented---that's standard.
- You don't have to pattern match to use `withRepr`
- Remove unnecessary paragraph prefixes like `Credit:` and `NOTE:`
- Move documentation on the default implementation to the method itself


langston-barrett:
Closing until I have time to revisit.

PR #185, opened 2025-03-19
ci: Move Haddocks to their own workflow

langston-barrett:
The setup I introduced in #183 was buggy. The `doc` job would fail on non-`master` branches, because it didn't have enough permissions to access the GitHub Pages "environment", even though if it were run it wouldn't attempt to upload to it.

Also, use a separate cache for documentation builds. This is appropriate, as we can and should build with `--disable-documentation` to save space and time in the `test` job, but *not* build with that flag for the docs. This discrepancy causes dependencies to be built differently, hence separate caches.

Return to checking Haddock well-formedness in `test`.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -1,0 +1,71 @@
+name: Documentation
+
+on:
+  push:
+    branches:
+    - master
+
+concurrency:
+  # Only run one of these at a time because they update the global GitHub Pages.
+  group: pages
+  cancel-in-progress: false
+
+jobs:
+  docs:
+    runs-on: ${{ matrix.os }}
+    name: Documentation
+    # This matrix is defined not in order to build multiple copies of the
+    # documentation in different configurations, but rather just to DRY the OS
+    # and GHC version. `runs-on` can't access variables from a top-level `env`
+    # block, only from `matrix`.
+    strategy:
+      matrix:
+        os: ["ubuntu-24.04"]
+        ghc: ["9.10.1"]
+    permissions:
+      contents: read
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+    - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


langston-barrett:
[Here](https://github.com/langston-barrett/parameterized-utils/actions/runs/13948914076) is a build demonstrating the `Documentation` workflow succeeding (it does so on my fork, because of the `push` trigger).

sauclovian-g:
Do you want me to hold off on tagging the release until you merge this? It shouldn't matter to the release as such but it seems like we'd want it if we ever start a stable branch from that release.

langston-barrett:
Yeah, might not be a bad idea to wait. Apologies for introducing issues just before a release!

.github/workflows/doc.yml
```diff
@@ -0,0 +1,71 @@
...
+          dist-newstyle
+        key: |
+          ${{ env.key }}-${{ github.ref }}
+        restore-keys: |
+          ${{ env.key }}-
+    - name: Build haddocks
+      run: >
+        cabal
+        haddock
+        --haddock-html-location='https://hackage.haskell.org/package/$pkg-$version/docs'
```
sauclovian-g, line 54: Is this the url wanted for uploading to GH Pages? I'd think you'd want a GH url...

sauclovian-g (approved):
All looks plausible, but I'm not really the expert on this stuff

.github/workflows/doc.yml
langston-barrett: This line ensures that links to other packages (e.g., `base`) redirect to their docs on Hackage. Without this line, all those links are broken because they end up pointing to a sub-URL of the GH pages site. Do you think this is worth a comment explaining this?

sauclovian-g: Yes. A lot of things about Haddock are fairly counterintuitive, this apparently being another...

langston-barrett: Done!

kquick (approved):

PR #188, opened 2025-03-20
doc: List some downstream projects in dev docs

langston-barrett:
By keeping track of this list, we can hopefully ensure that no important concerns get overlooked during the development process.

I'm sure there are projects that I'm not currently aware of that should be added to this list, please chime in with an entry for them!

### DIFF
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -1,5 +1,22 @@
 # Developer documentation
 
+## Downstream projects
+
+When evaluating breaking changes to the API, it can be helpful to consider the needs of major downstream projects (and even provide patches to migrate them).
+Here are a few, with brief descriptions of how they use this project:
+
+- [Crucible](https://github.com/GaloisInc/crucible):
+
+  - Crucible uses `Assignment`s to enforce variable-scoping constraints (i.e., the SSA invariant) on its control-flow graphs.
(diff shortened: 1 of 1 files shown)


doc/dev.md
```diff
@@ -1,5 +1,22 @@
...
 
+## Downstream projects
+
+When evaluating breaking changes to the API, it can be helpful to consider the needs of major downstream projects (and even provide patches to migrate them).
+Here are a few, with brief descriptions of how they use this project:
+
+- [Crucible](https://github.com/GaloisInc/crucible):
+
+  - Crucible uses `Assignment`s to enforce variable-scoping constraints (i.e., the SSA invariant) on its control-flow graphs.
+  - Crucible uses the classes from `TraversableFC` to manage its "syntax extensions", as well as its native expression and statement types.
```
RyanGlScott, line 11: Crucible's `TypeRepr` is probably worth a mention somewhere here.

RyanGlScott (approved):
It might worth be adding a mention to SAW as well, as it has a few interesting uses of `parameterized-utils` in its own right (e.g., [`TypeShape`](https://github.com/GaloisInc/saw-script/blob/6f4c278df4f468372e5cfe450f90ef35a74d7812/src/SAWScript/Crucible/MIR/TypeShape.hs#L52-L64)).

langston-barrett:
Thanks for the suggestions! Added.

PR #182, opened 2025-03-13
ci: Use `cabal-ghc-compat` action

langston-barrett:
I'm working on packaging up this stanza as a reusable composite action.

PR #191, opened 2025-03-26
ci: Simplify, move, and document the Cabal/GHC compatibility check

langston-barrett:
I was making a composite action for this (#182), but I realized that it could likely be simplified instead, to the point where such an action would be superfluous! I believe it still addresses the original concerns behind this check, but would appreciate confirmation from @kquick.

Added some more documentation along the lines of https://github.com/GaloisInc/.github/pull/13.


RyanGlScott (approved):

langston-barrett:
@kquick Penny for your thoughts?

PR #193, opened 2025-03-28
ci: Build packages and modules in parallel

langston-barrett:


langston-barrett:
This appears to speed a build where dependencies are unchanged but all local modules must be compiled (due to the same issue as https://github.com/GaloisInc/crucible/issues/1349) up from [~53s](https://github.com/GaloisInc/parameterized-utils/actions/runs/13977966039/job/39136313832) to [~32s](https://github.com/GaloisInc/parameterized-utils/actions/runs/14132882554/job/39597420385?pr=193). I imagine it would have a more dramatic positive impact on builds where the dependencies must also be compiled.

.github/workflows/test.yml
```diff
@@ -84,7 +84,22 @@ jobs:
       - name: Configure
         shell: bash
         # See doc/dev.md for development practices around warnings.
-        run: cabal configure --disable-documentation --enable-tests --ghc-options='-Werror=compat -Werror=default'
+        #
+        # See next step for `-j`.
+        run: cabal configure --disable-documentation --enable-tests --ghc-options='-Werror=compat -Werror=default -j'
```
RyanGlScott, line 89: I am leery about passing `-j` to GHC, as it is easy to oversubscribe your system if there are too many modules to build in parallel. (This is one of the motivations behind `Cabal`'s `--semaphore` flag.)

langston-barrett: Can you say more? [The docs](https://downloads.haskell.org/ghc/latest/docs/users_guide/using.html#ghc-flag-j-n) say:

> Perform compilation in parallel when possible. GHC will use up to ⟨N⟩ threads during compilation. If N is omitted, then it defaults to the number of processors.

*A priori*, it seems reasonable to me to use the same number of threads as processors, but perhaps Haskell builds tend to use a lot of RAM per module and that would become an issue? In that case, would something like `-j2` be appropriate, to allow a bit of parallelism without making it too likely to eat up all the RAM?

langston-barrett: [The GitHub docs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) say that a default Linux runner has 4 CPUs and 16GB RAM. So IIUC (and I'm not at all sure that I do), to oversubscribe would be to eat up all the RAM, which would require compiling 4 modules at once which each consume ~4 GB of RAM (minus the overhead for GHC and other system processes). Does my understanding match yours, and does that scenario sound likely? If so, do you think `-j2` (8 GB per module) would be ok?

RyanGlScott: The problem is described more in the motivation in [GHC proposal #540](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0540-jsem.rst). The problem is that build plans can be "wide" in some parts (i.e., lots of small packages that ought to be built in parallel with `ghc -j1`)) and "tall" in others (i.e., one large package whose modules should be built with `ghc -jN` for some large number `N`). It is difficult to pick a single number `N` to pass to all invocations of `ghc -jN` is a `cabal` build that will cover both wide and tall parts of build plans.

Rather than try to fiddle with `ghc -jN` manually, I would be more inclined to try the `--semaphore` approach, which uses a more principled approach to module-level parallelism.

langston-barrett: IIUC, passing `-j` in `cabal configure` like this will only apply to local packages (in this case, just p-u). Because it only applies to a single package, the "wide vs tall" problem doesn't arise. I agree that it would be a problem to pass `-j` for *all* packages, or to pass it to `cabal configure` if we invoked `cabal build` over multiple local packages at the same time.

langston-barrett: In other words, I think passing `--semaphore` to `cabal build` (of this package, not of the dependencies) would result in identical parallelism to `--ghc-options=-j` (as implemented here): Cabal would build a single package (p-u) in serial, making all available parallelism available to GHC.

RyanGlScott: > IIUC, passing `-j` in `cabal configure` like this will only apply to local packages (in this case, just p-u).

Does it say this somewhere in the `cabal` user guide? As far as I am aware, `--ghc-options` applies to _all_ invocations of GHC, not just ones related to local packages.

langston-barrett: When I run:
```
cabal configure --ghc-options='--foo'
```
I see
```
cat cabal.project.local

ignore-project: False

program-options
  ghc-options: --foo
```
And [the docs say](https://cabal.readthedocs.io/en/stable/cabal-project-description-file.html#cfg-section-program-options-program-options):

> [Program options](https://cabal.readthedocs.io/en/stable/config.html#program-options) can be specified once for all **local** packages by means of the program-options stanza.

(Emphasis mine)

So I think that it would just apply to local packages.

RyanGlScott: Ah, my mistake.

I think the "wide vs tall" problem would still exist, although it would be confined to parallelism between local packages. Granted,  that problem would likely be less severe than it would when considering all of the packages in a build plan. As such, my objection doesn't really apply—I think we could comfortably opt into some extra parallelism here.

Still, I'd love to just switch to `--semaphore` some day and avoid having to think about this so much :)

RyanGlScott (approved):

.github/workflows/test.yml
langston-barrett: All good, I think we've both learned some dark corners of Cabal today :smile: 

> I think the "wide vs tall" problem would still exist, although it would be confined to parallelism between local packages. Granted, that problem would likely be less severe than it would when considering all of the packages in a build plan. As such, my objection doesn't really apply—I think we could comfortably opt into some extra parallelism here.

Agreed, I think we're completely safe on this project in particular since it only consists of a single package, though.

> Still, I'd love to just switch to --semaphore some day and avoid having to think about this so much :)

:+1: :+1: :+1:

PR #195, opened 2025-04-09
ci: Use Dependabot for weekly bumps to pinned GitHub Actions

langston-barrett:
We have previously manually bumped the pinned versions of the actions used in our CI workflows. Let's automate this process using Dependabot. With this configuration, Dependabot will check for updates weekly, and open pull requests for each of them.

Benefits:

- Save maintainer time when updates are push-button
- See benefits from updated dependencies more quickly and reliably

Risks:

- Requires maintainers to review and merge more PRs
- May create noise for updates that aren't push-button and require some manual intervention

I'm not especially invested in this if people don't like it, I think that both the benefits and risks are fairly marginal.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "weekly"


RyanGlScott (approved):
I have mixed feelings about Dependabot, but if you're enthusiastic about using it, then I am on board.

langston-barrett:
> I have mixed feelings about Dependabot

I'd be interested in hearing about this sometime! I've enjoyed using it on my personal projects but don't have a ton of deep thoughts on the trade-offs.

RyanGlScott:
Configuring Dependabot can be painful for repos where you have [repo-specific secrets](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions), as Dependabot doesn't have permission to view them. Cryptol is one example of such a repo, which means that any time Dependabot opens a PR, the CI will break due to the secrets not being visible, and I have to manually force-push the PR's commit in order to convince GitHub to grant the appropriate permissions for repo secrets. There is almost certainly a better way to do this, granted, but I do get a distinct sense of dread every time I see Dependabot in my inbox as a result...

langston-barrett:
Gotcha. Well, we don't have repo-specific secrets here, so that shouldn't be a problem at least. Perhaps let's try it for a bit and see if it improves or worsens your sense of dread? :slightly_smiling_face: 

As I said above, I'm not really invested in this - if it has unforseen consequences I'd be very happy to learn my lesson and revert it. I could put a comment at the top of the configuration saying that it's provisional, and could even file an issue reminding us to reach a more final decision in a few weeks. Just let me know if any of that sounds helpful!

RyanGlScott:
I agree that using Dependabot in something as simple as `parameterized-utils`' CI workflow is fine. Let's try it out and see if my ennui lessens over time.

kquick:
Can we configure dependabot so that it only notifies us about bugfix versions?  Given that the actions currently work correctly, I don't know that feature additions or major version changes are of interest unless we also review/update/need that in this particular workflow.  Thus,  the main concern would be to get bugfixes.

langston-barrett:
> Can we configure dependabot so that it only notifies us about bugfix versions?

Not sure whether this is possible with Dependabot. However, we already are implicitly getting bugfixes, as we only pin the major (or in the case of `haskell-ci`, major.minor) versions of our actions. IIUC, this means that they fetch the latest versions of those actions when they run. So if that's what we want, we don't need to make any changes.

> I don't know that feature additions or major version changes are of interest 

These are actually the primary updates I had in mind! I'd like to get new versions of `haskell-ci` automatically, as there are a few feature ideas in the issue backlog that would be useful for any repo, such as support for `--semaphore`.

Additionally, bumping major versions occasionally is sometimes required to avoid deprecations, see e.g., https://github.com/GaloisInc/llvm-pretty/pull/151. I did similar PRs on a bunch of packages, would have loved to have had them be automated.

If it makes any difference, I'm happy to commit to reviewing and merging Dependabot PRs into any repo in which I enable it.

kquick (approved):
Thanks for clarifying the underlying concerns you are wanting to address.  I also appreciate your volunteering to be the sole maintainer, but I don't think we want to do that: it's not fair to you and it creates the type of bottleneck I think we'd like to try to avoid.

This seems like a reasonable repo to try this out on to evaluate its utility.

PR #196, opened 2025-04-11
Support GHC 9.12

### LINKED ISSUE #172: Support GHC 9.12
https://www.haskell.org/ghc/download_ghc_9_12_1.html

[Steps](https://github.com/GaloisInc/parameterized-utils/blob/master/doc/dev.md#ghc-versions):

- [ ] Allow the new version of `base` in the Cabal `build-depends`
- [ ] Run `cabal {build,test,haddock}`, bumping dependency bounds as needed
- [ ] Fix any new warnings from `-Wdefault`
- [ ] Add the new GHC version to the matrix in the GitHub Actions configuration
- [ ] Change the `doc` job to use the new GHC version
- [ ] Add the new version to the code that sets `GHC_NIXPKGS` in the CI config
- [ ] Add the new GHC version to the Cabal `tested-with` field
- [ ] Optionally follow the below steps to remove any old GHC versions

Upstream issues:

- https://github.com/hedgehogqa/haskell-hedgehog-classes/issues/63

langston-barrett:
Fixes #172

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
       matrix:
         os: [ubuntu-24.04]
         # See doc/dev.md for GHC version support policy
-        ghc: [9.10.1, 9.8.1, 9.6.1, 9.4.5, 9.2.2, 9.0.2, 8.10.7, 8.8.4, 8.6.5]
+        ghc: [9.12.2, 9.10.1, 9.8.1, 9.6.1, 9.4.5, 9.2.2, 9.0.2, 8.10.7, 8.8.4, 8.6.5]
       fail-fast: false
     uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v1
     with:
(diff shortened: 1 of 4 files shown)


sauclovian-g (approved):
LGTM

.github/workflows/test.yml
```diff
@@ -9,13 +9,16 @@ jobs:
...
-    uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v1.2
+    uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v1.4
     with:
       # See doc/dev.md for development practices around warnings.
       #
       # See Note [Parallelism] in the `haskell-ci` action for why `-j`.
-      configure-flags: --enable-tests --ghc-options='-Werror=compat -Werror=default -j'
+      #
+      # For `--allow-newer`, see
+      # https://github.com/hedgehogqa/haskell-hedgehog-classes/issues/63
```
RyanGlScott, line 21: I wonder if we should wait until there is a resolution to https://github.com/hedgehogqa/haskell-hedgehog-classes/issues/63 before proceeding here. I'm not opposed to using `--allow-newer` in case of an emergency, but then again, https://github.com/hedgehogqa/haskell-hedgehog-classes/issues/63 was only opened a day ago.

langston-barrett: There has been no movement on that issue in two weeks, how would you like to proceed?

I've submitted a PR: https://github.com/hedgehogqa/haskell-hedgehog-classes/pull/64

I built and ran tests locally and they passed.

RyanGlScott: I'll echo my usual sentiments that dependencies with uncertain maintainership make me nervous, so we should keep on eye on this going forward. That being said, the amount of hacks we need to do to work around this aren't too onerous, and they should be confined to `parameterized-utils` given that it only affects its test suite. I think we can move forward with this.

RyanGlScott (approved):

PR #208, opened 2025-09-30
Replace `lens` with `microlens{,-pro}`

### LINKED ISSUE #203: Drop dependency on `lens`?
`lens` is a big package with a lot of dependencies. Thus, anything that transitively depends on `lens` takes a long time to build. parameterized-utils doesn't really use many of the fancy features of `lens` either. Naively switching imports to use microlens (`Lens.Micro`) yields just a few errors:
```
src/Data/Parameterized/Fin.hs:39:1: error: [GHC-61948]
    Could not find module ‘Lens.Micro.Iso’.
    Perhaps you meant
      Lens.Micro.TH (needs flag -package-id microlens-th-0.4.3.17)
      Lens.Micro.Mtl (needs flag -package-id microlens-mtl-0.2.1.0)
      Lens.Micro (from microlens-0.4.14.0)
    Use -v to see a list of the files searched for.
   |
39 | import Lens.Micro.Iso (Iso', iso)
   | 
```
```
src/Data/Parameterized/Context/Unsafe.hs:1002:10: error: [GHC-76037]
    Not in scope: type constructor or class ‘Lens.Field1’
    NB: the module ‘Lens.Micro’ does not export ‘Field1’.
     |
1002 | instance Lens.Field1 (Assignment1 f t) (Assignment1 f u) (f t) (f u) where
     |
```
(and many more `Field*` errors)

I wonder if it might be worth dropping the dependency on `lens`, or at least hiding it behind a Cabal flag. We could deprecate the above functions / instances, and remove them in a later release.

langston-barrett:
This PR replaces our dependency on `lens` with dependencies on `microlens` and `microlens-pro`. The motivation is that `lens` has many, many (transitive) dependencies. This leads to:

- Slow build times (the `microlens` README claims that it compiles in 3.5s whereas `lens` takes  4m10s)
- Longer GHC upgrade latency (we have to wait until all of `lens`'s dependencies support a new GHC version)
- Smaller GHC support windows (our GHC support window is limited by that of all of our transitive dependencies)
- Supply chain risks such as lack of maintenance and even supply chain attacks (see recent NPM attacks)

With the most recent release of `microlens`, this doesn't even require removing functionality from parameterized-utils nor depending on `Internal` modules.

Fixes #203. IMO it's worth a follow-up where we deprecate the `Iso`s in `Fin` for a cycle and then remove our dependency on `microlens-pro`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
     uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v2
     with:
-      cache-key-prefix: v1
+      cache-key-prefix: v2
       # See doc/dev.md for development practices around warnings.
       #
       # See Note [Parallelism] in `haskell-ci.yml` for why `--ghc-options='-j'`


langston-barrett:
Oh, this is actually a breaking change... The problem is that [the `Field*` instances for `Assignment`](https://github.com/GaloisInc/parameterized-utils/blob/ebd59241828a97e06af6f8eeead0968a90108256/src/Data/Parameterized/Context/Unsafe.hs#L1002) are now instances of `Lens.Micro.Field*` instead of `Control.Lens.Field*`. So API consumers that use those instances will have to depend on `microlens` instead of `lens` to use `_1` &co.

I don't think this should be a huge problem, because I can't imagine a use of `Field*` that *isn't* just using `_1` etc. In particular, I don't think the `Field*` classes are often (or really, should be) used as constraints on function signatures or superclasses and the like.

For what it's worth, I found this when trying to change What4 to use `microlens`, which so far *also* seems quite feasible.

langston-barrett:
I was able to also port What4 to `microlens`: https://github.com/GaloisInc/what4/pull/323

sauclovian-g (commented):
I don't really have anything useful to add...

RyanGlScott (approved):
Indeed, I owe this a review. Unsurprisingly, not much to review here—nice to know that this in fact a drop-in replacement.

juhp:
Is the microlens lowerbound really necessary?: ie does it require 0.5?
(I can test myself, but it is so hard to tell often whether bounds are arbitrary or actually needed.)

(For example we don't have microlens 0.5 in Stackage Nightly [yet](https://github.com/commercialhaskell/stackage/issues/7851))

juhp:
(Nvm sorry sadly seems to be true the lower bound is indeed needed:

```
parameterized-utils/src/Data/Parameterized/Context/Unsafe.hs:1002:10: error: [GHC-76037]
    Not in scope: type constructor or class ‘Lens.Field1’
    Note: The module ‘Lens.Micro’ does not export ‘Field1’.
     |
1002 | instance Lens.Field1 (Assignment1 f t) (Assignment1 f u) (f t) (f u) where
     |          ^^^^^^^^^^^
```
)

PR #227, opened 2026-05-07
MapF: `SPECIALIZE` for `BinTree` functions

langston-barrett:
`MapF` is the only instance of `IsBinTree`. However, the GHC core showed that the dictionaries were not being unpacked. `MapF` is a hotspot in downstream consumers like Crucible, Macaw, and GREASE, so it's well worth optimizing.


src/Data/Parameterized/Map.hs
```diff
@@ -409,7 +438,6 @@ filterLt k m = fromMaybeS m (Bin.filterLt (compareKeyPair k) m)
 insert :: OrdF k => k tp -> a tp -> MapF k a -> MapF k a
 insert = \k v m -> seq k $ updatedValue (Bin.insert comparePairKeys (Pair k v) m)
 {-# INLINABLE insert #-}
--- {-# SPECIALIZE Bin.insert :: OrdF k => Pair k a -> MapF k a -> Updated (MapF k a) #-}
```
RyanGlScott, line 412: What is with this commented-out `SPECIALIZE` pragma (and the similar comments below)?

RyanGlScott (approved):
Huh, I have to admit that I find the design of this code strange, and I wonder if we wouldn't be better served by just removing the class in favor of writing the operations directly over `Pair`/`MapF`. That being said, I see that `IsBinTree` is part of the publicly exported API, so doing that could be disruptive. Using `SPECIALIZE` pragmas seems like a good approach for now.

src/Data/Parameterized/Map.hs
```diff
@@ -115,6 +115,30 @@ comparePairKeys :: OrdF k => Pair k a -> Pair k a -> Ordering
 comparePairKeys (Pair x _) (Pair y _) = toOrdering (compareF x y)
 {-# INLINABLE comparePairKeys #-}
 
+-- 'MapF' is the only 'IsBinTree' instance, so specialize each polymorphic
+-- 'BinTree' helper used below. This removes class-dictionary indirection
+-- through 'asBin'/'bin'/'size'/'tip' in the generated Core.
+{-# SPECIALIZE Bin.insert
```
kquick, line 121: You may want to use `SPECIALIZE INLINE` here (and for `delete` and `union` if not others).

src/Data/Parameterized/Map.hs
langston-barrett: No idea 🤷

langston-barrett:
> Huh, I have to admit that I find the design of this code strange, and I wonder if we wouldn't be better served by just removing the class in favor of writing the operations directly over Pair/MapF. That being said, I see that IsBinTree is part of the publicly exported API, so doing that could be disruptive. Using SPECIALIZE pragmas seems like a good approach for now.

My chain of thought exactly.

src/Data/Parameterized/Map.hs
langston-barrett: They're all `INLINABLE`, is there any particular advantage to using `SPECIALIZE INLINE` here?

langston-barrett:
Pushed another commit that achieves full speciailization via an `Internal` module. Some non-inlined internal `BinTree` helpers weren't getting specialized.

src/Data/Parameterized/Map.hs
kquick: According to https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/pragmas.html#specialize-pragma, if specialization occurs too early, it might not optimize (by which I assume inline), so one way to manage this is to specify the phase for inlining and/or specialization.  The `SPECIALIZE INLINE` is also separately described there with an example of achieving full unrolling of recursive application, and I would expect it would also mitigate against the phase concerns.

parameterized-utils.cabal
```diff
@@ -102,6 +102,7 @@ library
     Data.Parameterized.TraversableFC
     Data.Parameterized.TraversableFC.WithIndex
     Data.Parameterized.Utils.BinTree
+    Data.Parameterized.Utils.BinTree.Internal
```
RyanGlScott: Please put this in `other-modules` rather than `exported-modules`.

langston-barrett: Good call, thanks!

src/Data/Parameterized/Map.hs
langston-barrett: > According to https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/pragmas.html#specialize-pragma, if specialization occurs too early, it might not optimize (by which I assume inline), so one way to manage this is to specify the phase for inlining and/or specialization.

That page says:

> The main reason for using phase control on specialisations is so that you can write optimisation RULES that fire early in the compilation pipeline, and only then specialise the calls to the function. If specialisation is done too early, the optimisation rules might fail to fire.

I don't think there are any relevant `RULES` here.

> The `SPECIALIZE INLINE` is also separately described there with an example of achieving full unrolling of recursive application, and I would expect it would also mitigate against the phase concerns.

The example given has different type indices at different recursive unrollings, which I don't think applies here.

Just to be sure, I tried building with `SPECIALIZE INLINE` to have a peek at the core. It made no difference for `insert`, `delete`, nor `union`. Using it on all `BinTree` functions caused `Simplifier ticks exhausted`, which may be related to this warning from the docs:

> Warning
> 
> You can make GHC diverge by using `SPECIALISE INLINE` on an ordinarily-recursive function.

I'll leave it as-is for now.

Issue #113, opened 2021-05-19
Compilation failure when `unsafe-operations` is disabled


RyanGlScott:
If you build `parameterized-utils` with the `unsafe-operations` flag disabled, it will fail to compile:

```
$ cabal build -f-unsafe-operations
Build profile: -w ghc-8.10.4 -O1
In order, the following will be built (use -v for more details):
 - parameterized-utils-2.1.3.0.100 (lib) (file src/Data/Parameterized/Peano.hs changed)
Preprocessing library for parameterized-utils-2.1.3.0.100..
Building library for parameterized-utils-2.1.3.0.100..
[31 of 32] Compiling Data.Parameterized.Peano

src/Data/Parameterized/Peano.hs:326:1: error: [-Wincomplete-patterns, -Werror=incomplete-patterns]
    Pattern match(es) are non-exhaustive
    In an equation for ‘plusP’:
        Patterns not matched:
            ZRepr ZRepr
            ZRepr (SRepr _)
    |
326 | plusP (SRepr a) b = SRepr (plusP a b)
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

That particular error is easy enough to fix:

```diff
diff --git a/src/Data/Parameterized/Peano.hs b/src/Data/Parameterized/Peano.hs
index 55c97ca..7a5a5e6 100644
--- a/src/Data/Parameterized/Peano.hs
+++ b/src/Data/Parameterized/Peano.hs
@@ -323,6 +323,7 @@ plusP :: PeanoRepr a -> PeanoRepr b -> PeanoRepr (Plus a b)
 #ifdef UNSAFE_OPS
 plusP (PeanoRepr a) (PeanoRepr b) = PeanoRepr (a + b)
 #else
+plusP ZRepr     b = b
 plusP (SRepr a) b = SRepr (plusP a b)
 #endif
```

However, there are more errors waiting in the test suite:

<details>

```
$ cabal test -f-unsafe-operations
Build profile: -w ghc-8.10.4 -O1
In order, the following will be built (use -v for more details):
 - parameterized-utils-2.1.3.0.100 (test:parameterizedTests) (first run)
Preprocessing test suite 'parameterizedTests' for parameterized-utils-2.1.3.0.100..
Building test suite 'parameterizedTests' for parameterized-utils-2.1.3.0.100..
[1 of 7] Compiling Test.Context

test/Test/Context.hs:241:34: error:
[...]

PR #228, opened 2026-05-07
SmallArrayF: A `SmallArray#`-backed datastructure

langston-barrett:
As described in the Haddock, this is about trading `Assignment`'s amortized O(1) extend for `SmallArrayF`'s O(1) indexing and compactness for small-array, write-once, read-often scenarios, of which we have many in What4, Crucible, and GREASE.

This is a lot of unsafe code, including unsafe coercions and trusted rewrite rules. The testing strategy is:

- Property tests
- Property tests for tyupeclass instances
- Property tests for rule correctness - comparing LHS vs RHS
- Inspection tests for rules - making sure they produce the right Core

### DIFF
--- a/src/Data/Parameterized/SmallArrayF.hs
+++ b/src/Data/Parameterized/SmallArrayF.hs
@@ -1,0 +1,1091 @@
+{-|
+Module : Data.Parameterized.SmallArrayF
+Copyright : (c) Galois, Inc 2026
+Maintainer : Langston Barrett <langston@galois.com>
+
+A type-parameterized, fixed-size, heterogeneous container indexed by
+a type-level 'Data.Parameterized.Ctx.Ctx'.
+
+Compared with 'Data.Parameterized.Context.Assignment', 'SmallArrayF'
+trades \(O(1)\)-amortized 'extend' for \(O(1)\) indexing and \(O(n)\)
+'extend'/'update'.  Intended for small, read-heavy use cases such as
+machine register files.  For incremental construction, build an
+'Data.Parameterized.Context.Assignment' and convert via
+'fromAssignment' at the boundary.
+-}
+
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RoleAnnotations #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE Trustworthy #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UnboxedTuples #-}
+
+module Data.Parameterized.SmallArrayF
+  ( SmallArrayF
+    -- * Construction
+  , empty
+  , singleton
+  , from2
+  , from3
+  , from4
+  , from5
+  , from6
+  , from7
+  , from8
+  , from9
+  , generate
+  , generateA
+  , generateST
+  , generateIO
+    -- * Query
+  , (!)
+  , size
+    -- * Modification
+  , update
+  , adjust
+  , adjustM
+  , extend
+  , unextend
+    -- * Combine
+  , zipWithFC
+  , zipWithMFC
+  , zipWithMFCST
+  , zipWithMFCIO
+  , append
+    -- * Subarray
+  , take
+  , drop
+
+    -- * Traversal
+  , traverseFCST
+  , traverseFCIO
+  , itraverseFCST
+  , itraverseFCIO
+
+    -- * Conversion
+  , fromAssignment
+  , toAssignment
+  ) where
+
+import           Prelude hiding (drop, take)
+
+import           Control.DeepSeq (NFData(rnf))
+import           Control.Monad.IO.Class (MonadIO(liftIO))
+import           Control.Monad.ST (ST, runST, stToIO)
+import           Data.Functor.Identity (Identity(Identity, runIdentity))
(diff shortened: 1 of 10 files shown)


langston-barrett:
Weird, GHC 9.0 and 9.2 specifically are failing on this even when other versions compile it fine:
```
src/Data/Parameterized/SmallArrayF.hs:849:16: error:
Error:     • Couldn't match type: SmallArrayF f ctx0 -> m (SmallArrayF g ctx0)
                     with: forall (x :: Ctx k2). SmallArrayF f x -> m (SmallArrayF g x)
      Expected: (forall (x :: k2). f x -> m (g x))
                -> forall (x :: Ctx k2). SmallArrayF f x -> m (SmallArrayF g x)
        Actual: (forall (tp :: k). f tp -> m (g tp))
                -> SmallArrayF f ctx0 -> m (SmallArrayF g ctx0)
    • In the expression: traverseFC__
      In an equation for ‘traverseFC’: traverseFC = traverseFC__
      In the instance declaration for ‘TraversableFC SmallArrayF’
    • Relevant bindings include
        traverseFC :: (forall (x :: k2). f x -> m (g x))
                      -> forall (x :: Ctx k2). SmallArrayF f x -> m (SmallArrayF g x)
          (bound at src/Data/Parameterized/SmallArrayF.hs:[84](https://github.com/GaloisInc/parameterized-utils/actions/runs/25519707416/job/74900185195?pr=228#step:13:85)9:3)
    |
849 |   traverseFC = traverseFC__
    |                ^^^^^^^^^^^^
```

RyanGlScott:
I suspect that this might be due to limitations in how old versions of GHC handle subsumption when typechecking. You might try eta-expanding the definition (i.e., `traverseFC_ f xs = traverseFC__ f xs`) to see if that works around the issue.

langston-barrett:
Ugh, that breaks the `RULES`/inspection tests... Looks like it's time for CPP...

RyanGlScott:
You might also try `traverseFC_ f = traverseFC__ f`, which I think would also work around the issue (and be compatible with existing rewrite rules).

langston-barrett:
I appreciate the suggestion, unfortunately looks like even just eta-expanding the function argument leads to GHC floating out `lvl_*` variables that break the rules.

src/Data/Parameterized/SmallArrayF.hs
```diff
@@ -0,0 +1,1091 @@
...
+Module : Data.Parameterized.SmallArrayF
+Copyright : (c) Galois, Inc 2026
+Maintainer : Langston Barrett <langston@galois.com>
+
+A type-parameterized, fixed-size, heterogeneous container indexed by
+a type-level 'Data.Parameterized.Ctx.Ctx'.
+
+Compared with 'Data.Parameterized.Context.Assignment', 'SmallArrayF'
+trades \(O(1)\)-amortized 'extend' for \(O(1)\) indexing and \(O(n)\)
+'extend'/'update'.  Intended for small, read-heavy use cases such as
```
RyanGlScott, line 11: Is there a general guideline for what constitutes "small"? Say, no more than X elements?

src/Data/Parameterized/SmallArrayF.hs
```diff
@@ -0,0 +1,1091 @@
...
+    let !(I# i#) = Ctx.indexVal idx
+    liftIO (stToIO (writeMut# mut i# (toAny x)))
+  liftIO (stToIO (freezeMut mut))
+{-# SPECIALIZE generateIO ::
+      Size ctx
+   -> (forall tp. Index ctx tp -> IO (f tp))
+   -> IO (SmallArrayF f ctx) #-}
+
+unsafeMakeIndex :: Int -> Index ctx tp
+unsafeMakeIndex = unsafeCoerce
```
RyanGlScott, line 516: This coercion will only be safe if the `-funsafe-operations` `cabal` flag is enabled, as otherwise `Index` will be defined as a GADT that is not representationally equal to `Int`. Perhaps that's OK for now (building without `-funsafe-operations` is already kind of broken, see https://github.com/GaloisInc/parameterized-utils/issues/113), so we might want to leave a TODO about making this code work without `-funsafe-operations`.

src/Data/Parameterized/SmallArrayF.hs
```diff
@@ -0,0 +1,1091 @@
...
+
+------------------------------------------------------------------------
+-- Size
+
+-- | \(O(1)\).  The number of elements in the container.
+size :: SmallArrayF f ctx -> Size ctx
+size (SmallArrayF arr) =
+  -- SAFETY: 'Size' is a newtype over 'Int' and the array length equals
+  -- the number of positions in @ctx@ by construction.
+  unsafeCoerce (I# (size# arr))
```
RyanGlScott, line 527: Similar comments apply here, as `Size` will also be defined as a GADT when `-funsafe-operations` is not enabled.

test/Test/SmallArrayF.hs
```diff
@@ -0,0 +1,203 @@
...
+-- Build a SmallArrayF from a list of Some values by going through
+-- Assignment.
+mkSAF :: [Some f] -> Some (SAF.SmallArrayF f)
+mkSAF vals =
+  case go Ctx.empty vals of
+    Some a -> Some (SAF.fromAssignment a)
+  where
+    go :: Ctx.Assignment f ctx -> [Some f] -> Some (Ctx.Assignment f)
+    go acc []            = Some acc
+    go acc (Some x : xs) = go (Ctx.extend acc x) xs
```
RyanGlScott, line 37: I think this could also be defined by converting the `[Some f]` to an `Assignment` using [`fromList`](https://github.com/GaloisInc/parameterized-utils/issues/113), then appending the two `Assignments` using [`(<++>)`](https://hackage-content.haskell.org/package/parameterized-utils-2.3.0.0/docs/Data-Parameterized-Context.html#v:-60--43--43--62-).

src/Data/Parameterized/SmallArrayF.hs
langston-barrett: Excellent question. `SmallArray#` itself is undocumented in `GHC.Exts`, but the [`primitive`](https://hackage-content.haskell.org/package/primitive-0.9.1.0/docs/Data-Primitive-SmallArray.html) package has some documentation of it. To be quite honest, I can't say I fully understand it, but my impression is that the size only matters during construction/mutation, which is pretty limited in this interface. So I think `SmallArrayF` would actually be suitable for any amount of elements.

PR #233, opened 2026-06-11
`Data.Parameterized.Fin`: Add a `Hashable` instance

### LINKED ISSUE #232: `Data.Parameterized.Fin`: Add a `Hashable` instance
This could simply hash the underlying `NatRepr`.

RyanGlScott:
Fixes #232.

### DIFF
--- a/src/Data/Parameterized/Fin.hs
+++ b/src/Data/Parameterized/Fin.hs
@@ -34,6 +34,7 @@ module Data.Parameterized.Fin
   , fin0Absurd
   ) where
 
+import Data.Hashable (Hashable(..))
 import GHC.TypeNats (KnownNat)
 import Numeric.Natural (Natural)
 
@@ -49,6 +50,9 @@ instance Eq (Fin n) where
 
 instance Ord (Fin n) where
   compare i j = compare (finToNat i) (finToNat j)
+
+instance Hashable (Fin n) where
+  hashWithSalt salt (Fin i) = hashWithSalt salt i
 
 instance (1 <= n, KnownNat n) => Bounded (Fin n) where
   minBound = Fin (knownNat @0)
--- a/test/Test/Fin.hs
+++ b/test/Test/Fin.hs
@@ -11,11 +11,12 @@ module Test.Fin
   )
 where
 
+import           Data.Hashable (hashWithSalt)
 import           Numeric.Natural (Natural)
 
 import           Hedgehog
 import qualified Hedgehog.Gen as HG
-import           Hedgehog.Range (linear)
+import           Hedgehog.Range (linear, linearBounded)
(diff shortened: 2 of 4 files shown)


test/Test/Fin.hs
```diff
@@ -64,6 +73,11 @@ finTests =
             "minBound <= maxBound (2)"
             ((minBound :: Fin 2) <= (minBound :: Fin 2))
 
+      , testPropertyNamed
+          "Eq equality implies to hash equality"
```
RyanGlScott:
```suggestion
          "Eq equality implies hash equality"
```

src/Data/Parameterized/Fin.hs
```diff
@@ -50,6 +51,9 @@ instance Eq (Fin n) where
 instance Ord (Fin n) where
   compare i j = compare (finToNat i) (finToNat j)
 
+instance Hashable (Fin n) where
+  hashWithSalt salt (Fin i) = hashWithSalt salt i
```
langston-barrett, line 55: Would it make sense to make this `INLINEABLE`?

langston-barrett (approved):

src/Data/Parameterized/Fin.hs
RyanGlScott: Perhaps? I usually don't bother with marking small functions like this one with inline-related annotations, given that GHC is usually pretty eager to inline these anyway. Calling `ghc --show-iface dist-newstyle/build/aarch64-osx/ghc-9.6.7/parameterized-utils-2.3.0.0.99/build/Data/Parameterized/Fin.hi` suggests that `hashWithSalt` is already given an unfolding:

```
cb591713e2e6663da99f78ffff1ca611
  $fHashableFin_$chashWithSalt ::
    GHC.Types.Int -> Fin n -> GHC.Types.Int
  [HasNoCafRefs, TagSig: <TagProper>, LambdaFormInfo: LFReEntrant 2,
   Arity: 2, Strictness: <1!P(L)><1!P(A,1L)>, CPR: 1,
   Unfolding: Core: StableSystem <2,TrueFalse>
              \ @n :: GHC.Num.Natural.Natural
                (salt['GHC.Types.Many] :: GHC.Types.Int)
                (ds['GHC.Types.Many] :: Fin n) ->
              case ds of wild { Fin i irred i1 ->
              Data.Hashable.Class.$fHashableNatural_$chashWithSalt
                salt
                i1
                  `cast`
                (Data.Parameterized.NatRepr.Internal.N:NatRepr[0] <i>_N) }]
```

langston-barrett: Gotcha, I wasn't sure how much GHC did this on its own! Seems fine in that case, thanks for looking into it!

PR #235, opened 2026-06-11
`Data.Parameterized.Fin`: Add `Num` instance and related functions 

### LINKED ISSUE #231: `Data.Parameterized.Fin`: Add basic arithmetic support
Currently, `Fin n` doesn't offer many functions that allow you to compute interesting arithmetic-related results. I think it would be nice to add some operations that perform arithmetic modulo `n`. For example, I am particularly interested in:

```
addNat :: NatRepr n -> Fin n -> Fin n -> Fin n
mulNat :: NatRepr n -> Fin n -> Fin n -> Fin n
```

Other `Num`-related operations would also be nice as well.

RyanGlScott:
Fixes https://github.com/GaloisInc/parameterized-utils/issues/231.

In order to define these, I also needed to add a `withModLeq` helper function, which I put in a separate commit.

### DIFF
--- a/src/Data/Parameterized/Fin.hs
+++ b/src/Data/Parameterized/Fin.hs
@@ -23,15 +23,21 @@ that naturally has a fixed size bound of @n@.
 module Data.Parameterized.Fin
   ( Fin
   , mkFin
+  , mkFinModN
   , buildFin
   , countFin
   , viewFin
+  , finFromNatModN
   , finToNat
   , embed
   , tryEmbed
   , minFin
   , incFin
   , fin0Absurd
+  , addFinModN
+  , subFinModN
+  , mulFinModN
+  , negFinModN
   ) where
 
 import Data.Hashable (Hashable(..))
@@ -39,6 +45,7 @@ import GHC.TypeNats (KnownNat)
 import Numeric.Natural (Natural)
 
 import Data.Parameterized.NatRepr
+import Data.Parameterized.Some (Some(..))
 
 -- | The type @'Fin' n@ has exactly @n@ inhabitants.
 data Fin n =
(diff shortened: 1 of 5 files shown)


src/Data/Parameterized/Fin.hs
```diff
@@ -132,3 +181,24 @@ fin0Absurd =
       case plusComm x (knownNat @1) of
         Refl ->
           case addIsLeqLeft1 @1 @o @0 LeqProof of {})
+
+-- | Add two @'Fin' n@ values and reduce the result modulo @n@.
+addFin :: (1 <= n) => NatRepr n -> Fin n -> Fin n -> Fin n
+addFin n (Fin x) (Fin y) = mkFinModN n (addNat x y)
```
langston-barrett: It occurs to me that this is not the only thing one might call `addFin`. You could also define:
```haskell
addFin :: Fin n -> Fin m -> Fin (n + m)
```
It might be worth considering how we would name both of these functions if we wanted to include them both. Same goes for other operations.

langston-barrett (approved):

src/Data/Parameterized/Fin.hs
langston-barrett: One could also imagine:

- Parital addition `addMaybe :: Fin n -> Fin n -> Maybe (Fin n)`
- Saturating addition
- Unchecked addition

Though all of those would probably have some distinguishing feature in their naming scheme.

RyanGlScott: A fair point. How about suffixing these functions with `*ModN` (similar to `mkFinModN`)?

langston-barrett: Sounds good to me!

PR #239, opened 2026-06-16
Add `recipNatModN` and `withRecipModNat`

### LINKED ISSUE #237: `Data.Parameterized.Fin`: Add `recipFinModN` operation
In order to effectively use `Fin` values as [elements of a finite field](https://github.com/GaloisInc/what4/issues/411), it would be helpful to add an operation that efficiently computes the reciprocal (i.e., the modular inverse). That is, given `i :: Fin n`, compute `i^-1` such that `(i * i^1)` mod `n` equals `1`. I propose to give this operation the following name and type signature:

```hs
recipFinModN :: forall n. (1 <= n) => NatRepr n -> Fin n -> Maybe (Fin n)
```

This would be built on top of [`integerRecipMod#`](https://hackage-content.haskell.org/package/ghc-bignum-1.3/docs/GHC-Num-Integer.html#v:integerRecipMod-35-) from `ghc-bignum` under the hood. Note that `integerRecipMod#` is partial: if `gcd(i, n) > 1`, then it cannot compute a modular inverse. As such, `recipFinModN` will return `Nothing` in such scenarios. (Another scenario in which `integerRecipMod#` can fail is if the modulus equals `0`, but that cannot happen above due to the type-level `1 <= n` constraint.)

Note that in the context of finite fields, the modulus `n` will always be prime, which means that `gcd(i, n) = 1` for all `i` and therefore, the modular inverse will be defined for all `i`. That being said, it is unclear how best to enforce the requirement that `n` be prime at the type level. Barring that, the type signature above is (in my opinion) a reasonable compromise.

In order to define `recipFinModN`, it would also be helpful to define the following helper function over `NatRepr`s:

```hs
withRecipModNat ::
  forall n m a.
  (1 <= m) =>
  NatRepr n ->
  NatRepr m ->
  (forall r. (r + 1 <= m, Mod (n * r) m ~ 1) => NatRepr r -> a) ->
  Maybe a
```

RyanGlScott:
Fixes #237.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
     uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v2
     with:
-      cache-key-prefix: v5
+      cache-key-prefix: v7
       # See doc/dev.md for development practices around warnings.
       #
       # See Note [Parallelism] in `haskell-ci.yml` for why `--ghc-options='-j'`
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,8 +4,9 @@
 
 * Add `Hashable` and `Num` instances for `Fin`.
 * Add `mkFinModN`, `finFromNatModN`, `addFinModN`, `subFinModN`, `mulFinModN`,
-  and `negFinModN` to `Data.Parameterized.Fin`.
-* Add `modNat`, `modIsLeq`, and `withModLeq` to `Data.Parameterized.NatRepr`.
+  `negFinModN`, and `recipFinModN` to `Data.Parameterized.Fin`.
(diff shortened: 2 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
     uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v2
     with:
-      cache-key-prefix: v5
+      cache-key-prefix: v7
```
langston-barrett, line 20: Hmm, I wonder what's going on here... I really don't see these failures with other repos, and we revised the caching strategy to more closely model the pre-`haskell-ci`-workflow one, but it seems this still fails way too often due to caching issues...

langston-barrett (approved):

